PR #207, opened 2024-02-29
chore: update chess-library

Disservin:


Disservin:
should be fixed by https://github.com/Disservin/chess-library/pull/79 hopefully

PR #222, opened 2024-04-11
nelo bounds for SPRT

gahtan-syarif:
makes the elo0 and elo1 bounds for SPRT use normalized elo like in fishtest instead of logistic elo. this allows fastchess to directly use fishtest elo SPRT bounds.


Disservin:
gonna trust you on the maths on this one :D Probably in the future the llr should just take a Stats object `https://github.com/Disservin/fast-chess/blob/master/src/types/stats.hpp`.
Btw did fishtest have any tests associated with these functions, would make sense to add them.

gahtan-syarif:
for the penta llr i didn't take them from fishtest instead i used my own 5-nomial distribution code combined with the cpw llr formula that i used on my python elo and SPRT simulator here:
[pentanomial_elo_model.txt](https://github.com/Disservin/fast-chess/files/14952416/pentanomial_elo_model.txt)
the results are exactly the same as the fishtest one tho :)

PR #223, opened 2024-04-11
feat: replace trinomial stats with pentanomial stats

gahtan-syarif:
statistics functions are now overloaded so they can be called with pentanomial results. this would spit out pentanomial elo, error margins, draw ratio, score ratio, and LOS.

Edit: added better approximation for trinomial LOS since the last one by remi coulom doesnt include draw information and is made for convenience of calculation: https://www.talkchess.com/forum/viewtopic.php?t=51003&start=5 . Also replaced the entire phiinv function with the actual value for 95% confidence interval.

### DIFF
--- a/src/matchmaking/elo/elo.cpp
+++ b/src/matchmaking/elo/elo.cpp
@@ -47,8 +47,8 @@ double Elo::getError(int wins, int losses, int draws) noexcept {
     const double devD  = d * std::pow(0.5 - perc, 2.0);
     const double stdev = std::sqrt(devW + devL + devD) / std::sqrt(n);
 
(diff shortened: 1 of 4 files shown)


Disservin:
@gahtan-syarif this doesnt compile

gahtan-syarif:
> @gahtan-syarif this doesnt compile

@Disservin forgot to rename a variable, my bad, should compile now

gahtan-syarif:
Now that we have the penta stats is it better to display both wdl and penta or penta stats only? i guess wdl stats are more applicable when the games arent played in pairs (unequal white and black games). but as far as i can tell trinomial stats isnt displayed when -games are set to 1 right?

PR #225, opened 2024-04-12
Set default SPRT mode to pentanomial

gahtan-syarif:
also flip the pentanomial output from 2-0 to 0-2 to be more consistent with fishtest and Openbench.

### DIFF
--- a/src/matchmaking/tournament/roundrobin/roundrobin.cpp
+++ b/src/matchmaking/tournament/roundrobin/roundrobin.cpp
@@ -119,7 +119,7 @@ void RoundRobin::updateSprtStatus(const std::vector<EngineConfiguration>& engine
 
     const auto stats = result_.getStats(engine_configs[0].name, engine_configs[1].name);
     const auto llr = tournament_options_.report_penta ? sprt_.getLLR(stats.penta_WW, stats.penta_WD, stats.penta_WL, stats.penta_DD, stats.penta_LD, stats.penta_LL) 
-                                                      : sprt_.getLLR(stats.wins, stats.losses, stats.draws);
+                                                      : sprt_.getLLR(stats.wins, stats.draws, stats.losses);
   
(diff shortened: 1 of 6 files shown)


gahtan-syarif:
Screenshot 2024-04-12 130046
tested the output, calculations are correct :D

Disservin:
Man changes also need to run 

`make update_help`

Disservin:
Man changes also need to run 

`make update_help`

gahtan-syarif:
make update_help doesnt seem to work, reverted man change

src/matchmaking/tournament/roundrobin/roundrobin.cpp
```diff
@@ -118,7 +118,7 @@ void RoundRobin::updateSprtStatus(const std::vector<EngineConfiguration>& engine
     if (!sprt_.isValid()) return;
 
     const auto stats = result_.getStats(engine_configs[0].name, engine_configs[1].name);
-    const auto llr   = sprt_.getLLR(stats.wins, stats.draws, stats.losses);
+    const auto llr   = sprt_.getLLR(stats.penta_WW, stats.penta_WD, stats.penta_WL, stats.penta_DD, stats.penta_LD, stats.penta_LL);
```
Disservin, line 121: i think this doesnt work correctly, since it can be either cutechess or fastchess here..
something like
`const auto llr = tournament_options_.report_penta ? sprt_.getLLR(...) : sprt_.getLLR(...);`

Disservin (changes requested):

src/matchmaking/tournament/roundrobin/roundrobin.cpp
gahtan-syarif: ahh youre right, im assuming tournament_option_.report_penta is set to 0 when cutechess output is enabled?

gahtan-syarif: @Disservin fixed :D

PR #227, opened 2024-04-13
optional parameter for logistic elo SPRT bounds

gahtan-syarif:
since we're now using normalized elo bounds for sprt to be similar to fishtest and openbench, i wanted to add the option for users to instead use logistic elo bounds if they prefer that more. so we can add a parameter to the options to specify if they want to use normalized or logistic bounds like this:
```-sprt elo0=ELO0 elo1=ELO1 alpha=ALPHA beta=BETA bounds=(logistic/normalized)```
unfortunately idk how the options and parsing code in cli.cpp works, maybe you can help me out? @Disservin

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -249,6 +249,12 @@ void parseSprt(int &i, int argc, char const *argv[], ArgumentData &argument_data
             argument_data.tournament_options.sprt.alpha = std::stod(value);
         } else if (key == "beta") {
             argument_data.tournament_options.sprt.beta = std::stod(value);
+        } else if (key == "bounds") {
+            if (value == "logistic"){
+              argument_data.tournament_options.sprt.bounds = true;
+            } else if (value == "normalized"){
+              argument_data.tournament_options.sprt.bounds = false;
+            } else {OptionsParser::throwMissing("sprt", key, value);}
         } else {
             OptionsParser::throwMissing("sprt", key, value);
         }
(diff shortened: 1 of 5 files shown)


Disservin:
Adding a new option to sprt is rather easy, i.e. here https://github.com/Disservin/fast-chess/blob/master/src/cli/cli.cpp#L238, just add a new key and add the entry to the struct https://github.com/Disservin/fast-chess/blob/master/src/types/tournament_options.hpp#L29. Probably also good to have some value check not only key check
if value == logistic
else if value == normalized
else OptionsParser::throwMissing("sprt", key, value);

gahtan-syarif:
> Adding a new option to sprt is rather easy, i.e. here https://github.com/Disservin/fast-chess/blob/master/src/cli/cli.cpp#L238, just add a new key and add the entry to the struct https://github.com/Disservin/fast-chess/blob/master/src/types/tournament_options.hpp#L29. Probably also good to have some value check not only key check if value == logistic else if value == normalized else OptionsParser::throwMissing("sprt", key, value);

it worked! :D thank you

for the man page i think ill let you update it since i cant do make update_help

Disservin:
> > Adding a new option to sprt is rather easy, i.e. here https://github.com/Disservin/fast-chess/blob/master/src/cli/cli.cpp#L238, just add a new key and add the entry to the struct https://github.com/Disservin/fast-chess/blob/master/src/types/tournament_options.hpp#L29. Probably also good to have some value check not only key check if value == logistic else if value == normalized else OptionsParser::throwMissing("sprt", key, value);
> 
> it worked! :D thank you
> 
> for the man page i think ill let you update it since i cant do make update_help

btw cause windows? have you tried wsl?

PR #229, opened 2024-04-13
Display trinomial stats on fastchess output when -report penta=false

gahtan-syarif:
Output:
Screenshot 2024-04-13 224304

### DIFF
--- a/src/matchmaking/output/output_fastchess.hpp
+++ b/src/matchmaking/output/output_fastchess.hpp
@@ -21,13 +21,10 @@ class Fastchess : public IOutput {
 
     void printElo(const Stats& stats, const std::string& first, const std::string& second,
                   std::size_t current_game_count) override {
-        Elo elo;
-        if(report_penta_ == false){
-           elo = Elo(stats.wins, stats.losses, stats.draws);
-        } else {
-           elo = Elo(stats.penta_WW, stats.penta_WD, stats.penta_WL, stats.penta_DD,
+
+        const Elo elo(stats.wins, stats.losses, stats.draws);
+        const Elo eloPenta(stats.penta_WW, stats.penta_WD, stats.penta_WL, stats.penta_DD,
                      stats.penta_LD, stats.penta_LL);
-        }
 
         std::stringstream ss;
         ss << "Score of "   //
@@ -48,10 +45,10 @@ class Fastchess : public IOutput {
        
         if (report_penta_ == true){
            ss << "Elo difference: "   //
-              << elo.getElo()         //
+              << eloPenta.getElo()         //
               << ", "                 //
               << "nElo difference: "  //
-              << elo.getnElo()        //
+              << eloPenta.getnElo()        //
               << ", "                 //
               << "LOS: "              //
               << Elo::getLos(stats.penta_WW, stats.penta_WD, stats.penta_WL, stats.penta_DD,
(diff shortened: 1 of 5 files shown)


src/matchmaking/output/output_fastchess.hpp
```diff
@@ -34,38 +37,57 @@ class Fastchess : public IOutput {
            << "L - "        //
            << stats.draws   //
            << "D ["         //
-           << Elo::getScoreRatio(stats.penta_WW, stats.penta_WD, stats.penta_WL, stats.penta_DD,
-                                 stats.penta_LD, stats.penta_LL)  //
+           << Elo::getScoreRatio(stats.wins, stats.losses, stats.draws)  //
```
Disservin, line 40: doesnt this also need to be dynamic?

src/matchmaking/output/output_fastchess.hpp
```diff
@@ -8,18 +8,21 @@ namespace fast_chess {
...
         printSprt(sprt, stats);
-        printPenta(stats);
+        if (report_penta_ == true){printPenta(stats);}
         std::cout << "--------------------------------------------------\n";
     };
 
     void printElo(const Stats& stats, const std::string& first, const std::string& second,
                   std::size_t current_game_count) override {
-        const Elo elo(stats.penta_WW, stats.penta_WD, stats.penta_WL, stats.penta_DD,
+        const Elo elo(stats.wins, stats.losses, stats.draws);
```
Disservin, line 24: move it into the if scope?

src/matchmaking/output/output_fastchess.hpp
```diff
@@ -34,38 +37,57 @@ class Fastchess : public IOutput {
...
         if (sprt.isValid()) {
             std::stringstream ss;
 
-            ss << "LLR: " << std::fixed << std::setprecision(2)
-               << sprt.getLLR(stats.penta_WW, stats.penta_WD, stats.penta_WL, stats.penta_DD,
-                              stats.penta_LD, stats.penta_LL)
-               << " " << sprt.getBounds() << " " << sprt.getElo() << "\n";
+            if (report_penta_ == true){
+               ss << "LLR: " << std::fixed << std::setprecision(2)
+                  << sprt.getLLR(stats.penta_WW, stats.penta_WD, stats.penta_WL, stats.penta_DD,
```
Disservin, line 83: just have the getLLR instead the if branches the other stuff can be around it

Disservin (changes requested):

src/matchmaking/output/output_fastchess.hpp
gahtan-syarif: fixed with https://github.com/Disservin/fast-chess/pull/229/commits/59b00530813341486ef802b53d47efebe145cc3d

src/matchmaking/output/output_fastchess.hpp
gahtan-syarif: score ratio for pentanomial and wdl are the same when the number of games are even, since when penta_report=true stats are only printed every two games, the wdl function can be used

src/matchmaking/output/output_fastchess.hpp
gahtan-syarif: since its a declaration then if i move it into the if scope wouldnt the variable be limited to the if scope only? idk how to declare it separately outside the if scope

gahtan-syarif: nvm, fixed with https://github.com/Disservin/fast-chess/pull/229/commits/669363f34b42f6a061e44a1477861353ee5b1c68

PR #230, opened 2024-04-13
refactor: cleanup elo calculation

Disservin:


Disservin:
@gahtan-syarif thanks for all the additions, i'll refactor it like this so it's less of a mess ;)

gahtan-syarif:
no prob :)

PR #235, opened 2024-04-14
fix: fixes bug where when -games is set to 1, no statistic will be outputted

gahtan-syarif:
trinomial stats will now be automatically outputted when -games is set to 1


Disservin (changes requested):
i think this is the wrong place to do it, what if the user specified the report option before the games option? Then penta would be incorrectly turned off.
I think this has to go here https://github.com/Disservin/fast-chess/blob/master/src/matchmaking/tournament/tournament_manager.cpp#L24

gahtan-syarif:
@Disservin fixed :D

Disservin (approved):

PR #247, opened 2024-04-17
Makes sure score threshold is positive

gahtan-syarif:
makes sure the score threshold is converted to positive in case someone inputs a negative eval


Disservin:
what if someone wants a negative eval?
probably makes sense to warn the user once in cli.cpp but accept it?

Disservin:
Mh well I’d rather have it throw an error so that the user is aware of his mistake

gahtan-syarif:
@Disservin fixed 👍

Issue #248, opened 2024-04-17
bugs in pgn output


gahtan-syarif:
here is the pgn headers:
```
[Event "Fast Chess"]
[Site "?"]
[Round "0"]
[White "Engine2"]
[Black "Engine1"]
[Date "2024-04-17"]
[Result "1/2-1/2"]
[SetUp "1"]
[FEN "rn2kbnr/ppq2ppp/2p1p3/2Pp1b2/3P4/1QN2N2/PP2PPPP/R1B1KB1R w KQkq - hmvc 1; fmvn 7;"]
[GameDuration "00:00:30"]
[GameStartTime "2024-04-17T19:36:31 SE Asia Standard Time"]
[GameEndTime "2024-04-17T19:37:01 SE Asia Standard Time"]
[PlyCount "112"]
[TimeControl "10+0.1"]
```
the reported fen is invalid because its copied directly from the epd file, which means the hmvc and fmvn operators are not removed. also the gamestarttime and gameendtime is not accurate since the timestamp reported is in utc+0 while SE asia standard time is UTC+7, meaning the actual game time is off by 7 hours.

gahtan-syarif:
fixed with https://github.com/Disservin/fast-chess/commit/d6181bea7802b015e88d1d456c6f38c096282b7f and https://github.com/Disservin/fast-chess/pull/249

PR #249, opened 2024-04-17
output local time in PGN

gahtan-syarif:
fix time issue in https://github.com/Disservin/fast-chess/issues/248

### DIFF
--- a/src/util/date.hpp
+++ b/src/util/date.hpp
@@ -20,9 +20,9 @@ namespace time {
     struct tm buf {};
 
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


Disservin:
is this also what cutechess does?

gahtan-syarif:
> is this also what cutechess does?

wait ill check, but i think even if cutechess outputs the utc+0 we should still change it to this, cuz the local time is the correct one

gahtan-syarif:
@Disservin just checked, cutechess also outputs local time for its pgn, so this patch is consistent

PR #250, opened 2024-04-17
True EPD implementation

gahtan-syarif:
The current implementation of epd parsing treats all lines inside the epd file as fens and passes it directly to the engine. this is incorrect as the epd notation is not 100% the same as fen notation. for more information, see here: https://www.chessprogramming.org/Extended_Position_Description
cutechess also treats fens and epds differently, see here: https://github.com/cutechess/cutechess/blob/1071d84cf272bd7deca0964336bf02e367e2b22b/projects/lib/src/epdrecord.cpp#L22

fixes issue https://github.com/Disservin/fast-chess/issues/248

### DIFF
--- a/src/matchmaking/book/opening_book.cpp
+++ b/src/matchmaking/book/opening_book.cpp
@@ -35,8 +35,58 @@ void OpeningBook::setup(const std::string& file, FormatType type) {
         }
 
         openingFile.close();
-
-        book_ = epd;
+        
+        std::vector<std::pair<std::string, std::pair<int, int>>> fenData;
+    
+        for (const auto& line : epd) {
+            // Find the position of the first semicolon
+            size_t pos = line.find(';');
+            
+            // Find the position of the 4th space
+            size_t posspace = std::string::npos;
+            for (int i = 0, spaceCount = 0; i < line.size(); ++i) {
+                if (line[i] == ' ') {
+                    ++spaceCount;
+                    if (spaceCount == 4) {
+                        posspace = i;
+                        break;
+                    }
+                }
+            }
+    
+            // Extract first four FEN fields
+            std::string first_four_fen = line.substr(0, posspace);
+    
+            // Default values for hmvc and fmvn
+            int hmvc = 0, fmvn = 1;
+    
+            // Extract hmvc and fmvn values from the remaining parts
+            if (pos != std::string::npos) {
+                auto parts = split(line.substr(pos + 1), ';');
+                for (const auto& part : parts) {
+                    if (part.find("hmvc") != std::string::npos) {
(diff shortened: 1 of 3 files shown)


Disservin:
I think a split function already exists. Ideally I’d want this in my chess library instead though

gahtan-syarif:
> I think a split function already exists. Ideally I’d want this in my chess library instead though

should i pr this in your chess library repo? although im not familiar with the codebase there, or should you add it yourself?

Disservin:
I just added this here, https://github.com/Disservin/chess-library/pull/90. What then has to change is: https://github.com/Disservin/fast-chess/blob/master/src/matchmaking/match/match.cpp#L63. Where need a quick and dirty way to differentiate between epd and fen. I think the easiest is to check if a `;` is present?
`    data_ = MatchData(opening_.fen);` needs to be changed call getFen from the board method now.

gahtan-syarif:
> I just added this here, [Disservin/chess-library#90](https://github.com/Disservin/chess-library/pull/90). What then has to change is: https://github.com/Disservin/fast-chess/blob/master/src/matchmaking/match/match.cpp#L63. Where need a quick and dirty way to differentiate between epd and fen. I think the easiest is to check if a `;` is present? ` data_ = MatchData(opening_.fen);` needs to be changed call getFen from the board method now.

technically an epd can contain no ; if it doesnt have an optional operator. so maybe something like this?
```cpp
bool isFen(const std::string& line) {
    std::istringstream iss(line);
    std::string part;
    int count = 0;

    while (iss >> part) {
        count++;
    }

    // Check the number of parts
    if (count == 6 && line.find(';') == std::string::npos) {
        return true;
    } else {
        return false
    }
}
```
and shouldnt this be changed too? https://github.com/Disservin/fast-chess/blob/master/src/pgn/pgn_reader.cpp#L16

Disservin:
> and shouldnt this be changed too? https://github.com/Disservin/fast-chess/blob/master/src/pgn/pgn_reader.cpp#L24

that for reading a .pgn file these have FENs.

> technically an epd can contain no ; if it doesnt have an optional operator. so maybe something like this?


does it matter? if it has no optional operator then it's just  a shortened fen without move counters, no? I think the current setFen method might support that, if not just add `0 1`

gahtan-syarif:
> > and shouldnt this be changed too? https://github.com/Disservin/fast-chess/blob/master/src/pgn/pgn_reader.cpp#L24
> 
> thats for reading a .pgn file these have FENs.
> 
> > technically an epd can contain no ; if it doesnt have an optional operator. so maybe something like this?
> 
> does it matter? if it has no optional operator then it's just a shortened fen without move counters, no? I think the current setFen method might support that, if not just add `0 1`

ahh i see, no need for the part counter then

PR #251, opened 2024-04-18
update man

gahtan-syarif:
update man to include default values and latest information


gahtan-syarif:
idk why its failing, i already ran make update_man?

PR #253, opened 2024-04-18
Epd implementation and update chess library

gahtan-syarif:
@Disservin is this what you had in mind?

### DIFF
--- a/src/matchmaking/match/match.cpp
+++ b/src/matchmaking/match/match.cpp
@@ -59,7 +59,7 @@ void Match::addMoveData(const Player& player, int64_t measured_time_ms, bool leg
     data_.moves.push_back(move_data);
 }
 
-bool isFen(const std::string_view line) {
+bool isFen(const std::string line) {
     // Check the number of parts
     if (line.find(';') == std::string::npos) {
         return true;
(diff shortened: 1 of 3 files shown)


src/matchmaking/match/match.cpp
```diff
@@ -58,9 +58,22 @@ void Match::addMoveData(const Player& player, int64_t measured_time_ms, bool leg
     data_.moves.push_back(move_data);
 }
 
+bool isFen(const std::string line) {
```
Disservin, line 61: `const std::string& line`

src/matchmaking/match/match.cpp
```diff
@@ -58,9 +58,22 @@ void Match::addMoveData(const Player& player, int64_t measured_time_ms, bool leg
...
+    // Check the number of parts
+    if (line.find(';') == std::string::npos) {
```
Disservin, line 63: `return line.find(';') == std::string::npos;`

Disservin (changes requested):
yeah i think this should work

src/matchmaking/match/match.cpp
gahtan-syarif: ahh right hahah

src/matchmaking/match/match.cpp
gahtan-syarif: fixed 👍

src/matchmaking/match/match.cpp
gahtan-syarif: fixed 👍

Disservin (approved):

PR #256, opened 2024-04-19
fix pgn header for number of rounds

gahtan-syarif:
displayed rounds now start at 1 instead of 0.


Disservin:
At least in its current form it’s weird, the test says round is 1 when calling the builder but then the builder increments it to 2? huh? Do cutechess games start at 1 too?

gahtan-syarif:
> At least in its current form it’s weird, the test says round is 1 when calling the builder but then the builder increments it to 2? huh? Do cutechess games start at 1 too?

yes, cutechess pgns all start at round 1, and imo starting at round 0 doesnt make sense tbh. i guess internally its fine for it to be round 0 but in the pgn it should definitely have 1 added to to it which is why i only altered the displayed output. but if you want i can also make it initialized from 1 too?

Issue #259, opened 2024-04-19
order=random not working


gahtan-syarif:
@Disservin is it just me or is order=random not working for opening book? it just plays it sequentially,

here is my prompt: fast-chess.exe -openings file=openingss.epd format=epd order=random -engine cmd=Stockfish-dev-20240413-c55ae376.exe name=Engine1 -engine cmd=Stockfish-dev-20240413-c55ae376.exe name=Engine2 -each tc=10+0.1 -rounds 4 -concurrency 4 -resign movecount=1 score=600 -pgnout file=fastchess.pgn

gahtan-syarif:
https://github.com/disservin/fast-chess/blob/master/src/matchmaking/book/opening_book.cpp seems you forgot to add shuffle()?

Disservin:
oh yeah indeed

gahtan-syarif:
while youre on it i might want to add another bug that i just found, seems if you set the number of rounds to be greater than the number of openings in the book, the test seems to end prematurely, maybe something to do with how the book openings are recycled idk

PR #260, opened 2024-04-19
enable -order=random and fix issue where idx exceeds book size for epd books

gahtan-syarif:
fixes issue https://github.com/Disservin/fast-chess/issues/262 and https://github.com/Disservin/fast-chess/issues/259

### DIFF
--- a/src/matchmaking/book/opening_book.cpp
+++ b/src/matchmaking/book/opening_book.cpp
@@ -51,7 +51,7 @@ Opening OpeningBook::fetch() noexcept {
     static uint64_t opening_index = 0;
     const auto idx                = start_ + opening_index++;
     const auto book_size          = std::holds_alternative<epd_book>(book_)
-                                        ? (std::get<epd_book>(book_).size())
+                                        ? std::get<epd_book>(book_).size()
                                         : std::get<pgn_book>(book_).size();
 
     if (book_size == 0) {
--- a/src/matchmaking/book/opening_book.cpp
+++ b/src/matchmaking/book/opening_book.cpp
@@ -50,7 +50,7 @@ Opening OpeningBook::fetch() noexcept {
     static uint64_t opening_index = 0;
     const auto idx                = start_ + opening_index++;
     const auto book_size          = std::holds_alternative<epd_book>(book_)
-                                        ? std::get<epd_book>(book_).size()
+                                        ? (std::get<epd_book>(book_).size() - 1)
                                         : std::get<pgn_book>(book_).size();
 
     if (book_size == 0) {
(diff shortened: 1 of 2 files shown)


Disservin:
coulda just put the shuffle after both ifs

gahtan-syarif:
> coulda just put the shuffle after both ifs

fixed 👍

gahtan-syarif:
btw if i want to access the stats object in the openings cpp how do i go about doing it? im not familiar with how classes and that sort of stuff work

src/matchmaking/book/opening_book.cpp
```diff
@@ -42,21 +43,22 @@ void OpeningBook::setup(const std::string& file, FormatType type) {
...
+                                        ? (std::get<epd_book>(book_).size())
                                         : std::get<pgn_book>(book_).size();
 
     if (book_size == 0) {
         return {chess::constants::STARTPOS, {}};
     }
 
     if (std::holds_alternative<epd_book>(book_)) {
-        const auto fen = std::get<epd_book>(book_)[idx % std::get<epd_book>(book_).size()];
+        const auto fen = std::get<epd_book>(book_)[idx % (std::get<epd_book>(book_).size() - 1)];
```
Disservin, line 61: i don't get this..
a) why only subtract for epd books, and not for pgn books as well?
b) idx is "arbitrarily" large, modulo the book size will give an index which is [0, book_size), so why we subtracting one?

gahtan-syarif: @Disservin because its too big by 1, i tried debugging the book_.size by printing the function and everytime its larger than the actual opening size by 1, once i reduced it by one the tests no longer stopped prematurely. for the pgn the function outputs the correct amount of openings so i didnt reduce it

gahtan-syarif: issue https://github.com/Disservin/fast-chess/issues/262

Disservin: but then the issue is at some other place.. not here this is only a hack

gahtan-syarif: yeah, but i dont really know how the book_.size function works to calculate it so that its off by 1, and i checked the output of the pgns to see which openings are used and its correct for this fix, so maybe we can merge this first and do a refactor at some other point? cuz this bug is pretty critical

gahtan-syarif:
```
        while (safeGetline(openingFile, line)) {
            epd.emplace_back(line);
        }
```
could it be that this part of the code somehow add in an extra line? @Disservin

Disservin: perhaps yea, maybe a 
```cpp
        while (safeGetline(openingFile, line)) {
            if (!line.empty()) epd.emplace_back(line);
        }
```

gahtan-syarif: alright ill try it and see the output first

Disservin: fixed it locally for me i think

gahtan-syarif: yep fixed it for me too, thank you

PR #264, opened 2024-04-20
fix: opening order now continues from last saved game

gahtan-syarif:
with this patch, when games are loaded from config file, the opening order is automatically offset so that the next opening continues from the last game before it was saved. this ensures that openings dont get repeated.

### DIFF
--- a/src/matchmaking/book/opening_book.hpp
+++ b/src/matchmaking/book/opening_book.hpp
@@ -17,8 +17,8 @@ class OpeningBook {
     explicit OpeningBook(const options::Opening& opening);
 
     /// @brief Fisher-Yates / Knuth shuffle
-    void shuffle() noexcept {
-        constexpr auto shuffle = [](auto& vec) {
+    void shuffle() {
+        const auto shuffle = [](auto& vec) {
             for (std::size_t i = 0; i + 2 <= vec.size(); i++) {
                 auto rand     = random::mersenne_rand();
                 std::size_t j = i + (rand % (vec.size() - i));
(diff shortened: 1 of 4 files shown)


Disservin:
eh that is gonna be a "difficult", since a tournament is created and then the book and only afterwards the results are set. You will need to add a method to openingbook and then call that from `setResults` in `BaseTournament`

gahtan-syarif:
hmm i see, let my try it out

gahtan-syarif:
nvm idk what im doing
is there an easier method? maybe just use the stats object from stats.hpp?

Disservin:
here pushed some stuff for here, i think that should give you access to it

gahtan-syarif:
wait why delete the games_ variable?

gahtan-syarif:
finished 👍 thanks

src/matchmaking/book/opening_book.hpp
```diff
@@ -14,7 +14,7 @@ namespace fast_chess {
 class OpeningBook {
    public:
     OpeningBook() = default;
-    explicit OpeningBook(const options::Opening& opening);
+    explicit OpeningBook(const options::Opening& opening, const options::Tournament& tournament);
```
Disservin, line 17: passing opening is useless cause its on the torunamentr already

gahtan-syarif: fixed 👍

gahtan-syarif:
@Disservin any issues with this one?

gahtan-syarif:
ahh merge conflicts

PR #275, opened 2024-04-21
feat: implement pgn reader limit for plies

### LINKED ISSUE #267: plies option does not work
specifying the number of plies to use from a pgn book does not work, it still uses the maximum number of plies available

Disservin:
fixes #267


gahtan-syarif:
doesnt seem to work for me

Disservin:
> doesnt seem to work for me

ah true actually, i havent changed the parameters in the function call

PR #276, opened 2024-04-21
fix: reimplement ratinginterval change

### LINKED ISSUE #266: ratinginterval option does not work
ratinginterval option has no effect on the displayed output.

Disservin:
something like this `tournament_options_.report_penta ? round_id : game_id;` maybe has something weird to it

fixes #266


gahtan-syarif:
i think round_id increments from 0 tho while game_id increments from 1, so maybe round_id+1?

gahtan-syarif:
works 👍 but youll also want to change the default ratinginterval value from 10 to 1 in tournament_options.hpp

Disservin:
well i think 10 is a reasonable value to not spam the user.

PR #277, opened 2024-04-21
fix: opening order now starts from last saved game

gahtan-syarif:
@Disservin fixed the merge conflicts


gahtan-syarif:
also fix issue where order=sequential is not working

Disservin:
btw i wasn't so sure how the correct thing would look like for this, but i guess this will do

PR #282, opened 2024-04-23
fix ratinginterval condition

gahtan-syarif:
fixes issue where output is printed per game instead of per pair when games=2 and report penta=false


gahtan-syarif:
nvm 😅

PR #283, opened 2024-04-23
add cutechess hyperlink to readme

gahtan-syarif:
we reference cutechess output in the readme so i think it makes sense to link to their page as well so people know what it is


Disservin:
id honstely leave that just out

Issue #285, opened 2024-04-25
timemargin support


gahtan-syarif:
do we need this? OB seems to use default timemargin of 250 with cutechess, but then again fast-chess experiences much less timelosses than cutechess so idk if this option is necessary or not

PR #286, opened 2024-04-25
feat: add -timemargin option

gahtan-syarif:
i think this works, dont really know how to test it since fast-chess rarely if ever experiences timelosses lol. i dont know if we need this so your call to merge this or not.

fix issue https://github.com/Disservin/fast-chess/issues/285


src/matchmaking/player.hpp
```diff
@@ -30,17 +30,21 @@ class Player {
...
         time_control_.time -= elapsed_millis;
 
-        if (time_control_.time < 0) {
+        if (time_control_.time < -timemargin) {
             return false;
         }
 
+        if (time_control_.time < 0) {
+            time_control_.time = 0;
+        }
```
Disservin, line 46: i dont really get this, why set it to 0?

gahtan-syarif: so it prevents negative time which i feel might lead to weird behavior,
and the way i see it, the timemargin is just an extra time after the actual time control that the engine is alloted, so after the move is made it shouldnt be part of the calculation for the final time after increment is added, 
say an engine exceeded the timelimit by 250ms with a 1 second increment, this way the final time after the move should be 1 second instead of 750ms

gahtan-syarif:
theres merge conflicts, ill resolve it first

PR #288, opened 2024-04-26
feat: add epdout

### LINKED ISSUE #284: -epdout addition?
cutechess has -epdout option to output the final fen of each game in an epd file, i think this one is pretty useful for constructing opening books when combined with draw adjudication/resign

Disservin:
fixes #284

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -201,6 +201,16 @@ void parsePgnOut(int &i, int argc, char const *argv[], ArgumentData &argument_da
     }
 }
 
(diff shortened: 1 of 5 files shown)


gahtan-syarif:
hmm doesnt seem to work for me, if i specify it alongside pgnout then it outputs a single fen at the end of the pgn file, and if i use the epdout option alone it causes all the engines to timeloss and end the match prematurely.

also warning from compiler than round_id param in the function is unused

Disservin:
ah right forgot to create multiple files

gahtan-syarif:
works 👍

PR #290, opened 2024-04-26
feat: add -timemargin option

gahtan-syarif:
I set the time control to 0 if its below 0 so it prevents negative time which i feel might lead to weird behavior,
and the way i see it, the timemargin is just an extra time after the actual time control that the engine is alloted, so after the move is made it shouldnt be part of the calculation for the final time after increment is added,
say an engine exceeded the timelimit by 250ms with a 1 second increment, this way the final time after the move should be 1 second instead of 750ms

fix https://github.com/Disservin/fast-chess/issues/285

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -404,6 +404,10 @@ void parseRatinginterval(int &i, int argc, char const *argv[], ArgumentData &arg
     parseValue(i, argc, argv, argument_data.tournament_options.ratinginterval);
 }
 
+void parseTimemargin(int &i, int argc, char const *argv[], ArgumentData &argument_data) {
+    parseValue(i, argc, argv, argument_data.tournament_options.timemargin);
+}
+
 void parseSRand(int &i, int argc, char const *argv[], ArgumentData &argument_data) {
     parseValue(i, argc, argv, argument_data.tournament_options.seed);
 }
@@ -518,6 +522,7 @@ OptionsParser::OptionsParser(int argc, char const *argv[]) {
     addOption("games", parseGames);
     addOption("rounds", parseRounds);
     addOption("ratinginterval", parseRatinginterval);
+    addOption("timemargin", parseTimemargin);
     addOption("srand", parseSRand);
     addOption("version", parseVersion);
     addOption("-version", parseVersion);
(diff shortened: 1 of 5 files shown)


gahtan-syarif:
btw i removed the randomseed from the json so that when loaded from config file it wouldnt generate a new seed instead it would continue from the old one

Disservin:
> btw i removed the randomseed from the json so that when loaded from config file it wouldnt generate a new seed instead it would continue from the old one

why would it? we are not saving a seed, but rather the boolean if randomseed is turned on/off?

gahtan-syarif:
> > btw i removed the randomseed from the json so that when loaded from config file it wouldnt generate a new seed instead it would continue from the old one
> 
> why would it? we are not saving a seed, but rather the boolean if randomseed is turned on/off?

if the randomseed boolean is saved to be true then wouldnt it generate a new random seed when the game is loaded from a config file?

Disservin:
why not ? 

https://github.com/Disservin/fast-chess/blob/master/src/matchmaking/tournament/tournament_manager.cpp#L12

gahtan-syarif:
I feel the random seed should be generated only once when the game is initially started and when the game is saved and then loaded it would still use that same seed, this prevents the same openings from ever repeating, which is also why i issued this commit https://github.com/Disservin/fast-chess/pull/277

gahtan-syarif:
added the new epd option to the json

Disservin:
I see but in my mind the config should be an exact replica of the options you specified over the cli.

gahtan-syarif:
> I see but in my mind the config should be an exact replica of the options you specified over the cli.

maybe we can make it so that the randomseed option is still saved to the json but wouldnt generate a new seed when loaded from a config file by adding an extra condition to the random seed geneartor to only activate when round_id == 0? @Disservin

Disservin:
apparently i didnt send, i think that sounds good

gahtan-syarif:
is there a way we can access the round_id variable in the TournamentManager class? if not round_id then maybe match_count_ from tournament.hpp

gahtan-syarif:
> is there a way we can access the round_id variable in the TournamentManager class? if not round_id then maybe intital_matchcount_ from tournament.hpp

gahtan-syarif:
> is there a way we can access the round_id variable in the TournamentManager class? if not round_id then maybe match_count_ from tournament.hpp

nvm found another way, i think this should work now. if the seed from the config is the default seed then the seed randomizer would activate and randomize it, if the seed from the config is not the default seed then it wouldnt randomize it

gahtan-syarif:
done

gahtan-syarif:
i think this can be merged now? @Disservin

PR #298, opened 2024-04-27
refactor: make timemargin applicable to individual engines

Disservin:


Disservin:
@gahtan-syarif it isnt that difficult, is it?^^

gahtan-syarif:
lol nvm what i said then :D

gahtan-syarif:
btw i noticed the timemargin is no longer stored in the tournament_options header? where is it stored alongside all the other engine config options

Disservin:
there are two main config objects, tournament and engine ones
https://github.com/Disservin/fast-chess/blob/master/src/main.cpp#L17

engineconfig is also in types see:
https://github.com/Disservin/fast-chess/blob/master/src/types/engine_config.hpp

Issue #302, opened 2024-04-28
timeouts for fixed time control


gahtan-syarif:
timeouts dont happen under fixed time control, this is my prompt
```
fast-chess.exe -openings file=openingss.epd format=epd -engine cmd=Stockfish-dev-20240413-c55ae376.exe name=Engine1 -engine cmd=Stockfish-dev-20240413-c55ae376.exe name=Engine2 -each st=0.1 timemargin=0 -rounds 4 -concurrency 4 -resign movecount=1 score=600 -pgnout file=fastchess.pgn -ratinginterval 1 -log file=log
```

I intentionally slowed down my computer hoping for a timeloss and this is the resulting pgn:
```
[Event "Fast Chess"]
[Site "?"]
[Round "1"]
[White "Engine2"]
[Black "Engine1"]
[Date "2024-04-28"]
[Result "1/2-1/2"]
[SetUp "1"]
[FEN "rn1qkb1r/3ppp1p/b4np1/2pP4/8/2N5/PP2PPPP/R1BQKBNR w KQkq - 3 8"]
[GameDuration "00:00:12"]
[GameStartTime "2024-04-28T07:19:47 SE Asia Standard Time"]
[GameEndTime "2024-04-28T07:19:59 SE Asia Standard Time"]
[PlyCount "87"]
[TimeControl "0"]

1. e4 {+0.68/12, 0.232s} Bxf1 {-0.74/12, 0.414s} 2. Kxf1 {+0.57/11, 0.202s}
d6 {-0.66/12, 0.156s} 3. a4 {+0.70/11, 0.115s} Bg7 {-0.55/12, 0.101s}
4. g3 {+0.72/13, 0.107s} Qd7 {-0.60/11, 0.112s} 5. Kg2 {+0.73/10, 0.111s}
O-O {-0.73/9, 0.159s} 6. Nf3 {+0.77/11, 0.100s} Na6 {-0.73/11, 0.102s}
7. Nd2 {+0.99/12, 0.103s} Rfb8 {-0.91/12, 0.119s} 8. Nc4 {+1.03/12, 0.104s}
Rb4 {-0.98/12, 0.105s} 9. Qe2 {+1.00/12, 0.109s} h5 {-0.97/13, 0.119s}
10. h3 {+1.01/12, 0.110s} e6 {-1.08/12, 0.117s} 11. a5 {+1.26/12, 0.166s}
[...]

PR #303, opened 2024-04-28
refactor: move logging out of read loop

Disservin:


Disservin:
perhaps helps https://github.com/Disservin/fast-chess/issues/302

PR #320, opened 2024-04-29
fix: reformat pgn date tag and change pgn tag order to follow standard

gahtan-syarif:
correct order is: event-site-date-round-white-black-result. Also reformat date to YYYY.MM.DD. cutechess also follows this standard. see below:

Screenshot 2024-04-29 072444


Disservin:
Just change the format here https://github.com/Disservin/fast-chess/blob/a085af77f5d75d9ff48fa6ad43e669fda1184e92/src/types/match_data.hpp#L60

gahtan-syarif:
> Just change the format here
> 
> https://github.com/Disservin/fast-chess/blob/a085af77f5d75d9ff48fa6ad43e669fda1184e92/src/types/match_data.hpp#L60

fixed 👍

PR #326, opened 2024-04-30
fix: revise updateTime

gahtan-syarif:
i feel this part of the code:
```
        if (tc.time_left == 0) {
            return true;
        }
```
is unnecessary and might lead to weird behavior where if the timeleft ever reaches zero, the updateTime function would perpetually return true and timeout would never happen and time increment wont be added

### DIFF
--- a/src/matchmaking/player.hpp
+++ b/src/matchmaking/player.hpp
@@ -84,7 +84,9 @@ class Player {
             auto white = stm == chess::Color::WHITE ? time_control_ : enemy_tc;
             auto black = stm == chess::Color::WHITE ? enemy_tc : time_control_;
 
-            input << " wtime " << white.time_left << " btime " << black.time_left;
+            if (time_control_.time != 0) {
+               input << " wtime " << white.time_left << " btime " << black.time_left;
+            }
 
             if (time_control_.increment != 0) {
                 input << " winc " << white.increment << " binc " << black.increment;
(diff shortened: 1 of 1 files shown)


src/matchmaking/player.hpp
```diff
@@ -34,10 +36,6 @@ class Player {
     // Returns false if the time control has been exceeded.
     [[nodiscard]] bool updateTime(const int64_t elapsed_millis) {
         auto &tc = time_control_;
-        if (tc.time_left == 0) {
-            return true;
```
Disservin, line 38: i right now wonder why i had return true here instead of false

src/matchmaking/player.hpp
```diff
@@ -12,8 +12,10 @@ class Player {
         : engine(uci_enigne), time_control_(uci_enigne.getConfig().limit.tc) {
         if (time_control_.fixed_time != 0) {
             time_control_.time_left = time_control_.fixed_time;
-        } else {
+        } else if (time_control_.time != 0) {
             time_control_.time_left = time_control_.time;
+        } else {
+            time_control_.time_left = INT64_MAX;
```
Disservin, line 18: id prefer `std::numeric_limits<std::int64_t>::max()`

gahtan-syarif: ahh ok

gahtan-syarif: updated 👍

src/matchmaking/player.hpp
gahtan-syarif: i think you added it in so that when the user is running a fixed depth/node test it wouldnt immediately timeloss

Disservin: As thanks that sounds reasonable, I was just wondering if this was actually helping in avoiding timelosses. I will later check since when it’s there

Issue #217, opened 2024-04-02
Affinity parser does not work for certain linux systems


Disservin:
The parser depends on the existence of `processor`, `core id` and `physical id` in `/proc/cpuinfo` it should however also be capable to parse it with only `processor` being present, this can be the case for non hyperthreaded systems.

PR #327, opened 2024-04-30
modified parser to work without physical id and core id

NekSaikou:
https://github.com/Disservin/fast-chess/issues/217

### DIFF
--- a/src/affinity/cpuinfo/cpuinfo_posix.hpp
+++ b/src/affinity/cpuinfo/cpuinfo_posix.hpp
@@ -22,6 +22,20 @@ inline CpuInfo getCpuInfo() {
     int coreId      = -1;
     int physicalId  = -1;
 
+    bool procIdFound     = false;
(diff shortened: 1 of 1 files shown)


Disservin:
I think this is indeed a bit more fail safe, though the logical is slightly wrong because it stores it as 1 core with multiple processors, whereas with hyper threading we have two processors per core and with non hyper threading 1 processor per core, this will end up with (possibly) more than 1/2 processors per core. Which is right now no big deal in the code, but the code for the affinity setting is actually a bit flawed.
https://github.com/Disservin/fast-chess/commit/423bb14dcb7c2803cd776316790a1fdff93e4192#diff-12ba2bd0a2975e9599c30836516443df104515698a92dc945d8146f0ab4dd02fR100
Idk here im making a vector of 1 cpu, and i think with the old core weird stuff would happen with Threads > 1 (engine settings)...

Issue #328, opened 2024-04-30
movestogo doesnt work


gahtan-syarif:
i specified tc=40/10+0.1 but as can be seen in the log file below, movestogo is stuck at 40 for the entire match.
[log.txt](https://github.com/Disservin/fast-chess/files/15160993/log.txt)

On top of that, ideally, fastchess can also do something like tc=40/10+0.1:20+0.2 where it stays at 10+0.1 tc for the 40 moves and then resets and use 20+0.2 for the rest of the game. idk if this feature is already implemented in fastchess (this feature isnt available in cutechess as well iirc so ig its fine if it isnt implemented but be a lot cooler if it did)

PR #330, opened 2024-04-30
fix: movestogo

Disservin:
fixes https://github.com/Disservin/fast-chess/issues/328

### DIFF
--- a/src/matchmaking/player.hpp
+++ b/src/matchmaking/player.hpp
@@ -17,6 +17,8 @@ class Player {
         } else {
             time_control_.time_left = std::numeric_limits<std::int64_t>::max();
         }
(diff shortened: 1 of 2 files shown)


gahtan-syarif:
in here: https://github.com/Disservin/fast-chess/blob/ff150fd129c6b6841ae209478e5dac1734f31e1d/src/matchmaking/player.hpp#L111 it should be moves_left

gahtan-syarif:
and in here https://github.com/Disservin/fast-chess/blob/ff150fd129c6b6841ae209478e5dac1734f31e1d/src/matchmaking/player.hpp#L20 i think you should include moves_left=moves

PR #331, opened 2024-04-30
chore: bump chess

Disservin:


Disservin:
made the `uci::uciToMove` more failsafe with very bad input to not have ub/crash.

Issue #332, opened 2024-04-30
feature: specify tc after movestogo


Disservin:
add something like tc=40/10+0.1:20+0.2 where it stays at 10+0.1 tc for the 40 moves and then resets and use 20+0.2 for the rest of the game. idk if this feature is already implemented in fastchess (this feature isnt available in cutechess as well iirc so ig its fine if it isnt implemented but be a lot cooler if it did)

PR #334, opened 2024-05-01
feat: add support for overtime format

gahtan-syarif:
complete TC format is now something like this: 40/60+0.6:20/10+0.1

already tested it, works great
satisfies issue https://github.com/Disservin/fast-chess/issues/332

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -62,11 +62,19 @@ TimeControl parseTc(const std::string &tcString) {
     TimeControl tc;
 
     std::string remainingStringVector = tcString;
+    std::string secondtc, remainingStringVector2;
     const bool has_moves              = str_utils::contains(tcString, "/");
     const bool has_inc                = str_utils::contains(tcString, "+");
+    const bool has_secondtc           = str_utils::contains(tcString, ":");
 
+    if (has_secondtc) {
+        const auto fulltc     = str_utils::splitString(remainingStringVector, ':');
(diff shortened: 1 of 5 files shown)


gahtan-syarif:
done

gahtan-syarif:
done

Disservin:
well mh i dont like the added complexity and given that no one really uses movestogo anyway..

gahtan-syarif:
> well mh i dont like the added complexity and given that no one really uses movestogo anyway..

i disagree, say if you want to emulate a FIDE tournament match
Screenshot_2024-04-12_030726
theres no way to do it in cutechess, with this patch fastchess offers that functionality.
also with this patch the TC now conforms to the PGN time control format
i think this patch would give us an edge over cutechess, same like penta sprt

Disservin:
well sure but who does that anyway, the scope initially was to create something which is faster than cutechess and can be compiler more easily

the spec stuff can be it's own pr

gahtan-syarif:
> well sure but who does that anyway, the scope initially was to create something which is faster than cutechess and can be compiler more easily
> 
> the spec stuff can be it's own pr

idk i think the complexity added isnt so much that its no longer understandable, plus its still backwards compatible with the old time format so i dont really see the problem,
yeah i guess this is first and foremost a cutechess alternative but if we can add new features and improve upon cutechess even further i dont see why not honestly, especially since we can now support the official FIDE time format with this feature

PR #337, opened 2024-05-01
refactor: stop tournament when the engine cant be started

Disservin:


gahtan-syarif:
btw does this mean we're removing -recover option?

Disservin:
Basically, I figured the recover option wasnt really implemented in the way cutechess has it, I think cutechess recovers the engine and continues with playing the same game, whereas in fastchess this would restart the entire game over and over.

PR #341, opened 2024-05-02
fix: prevent engine names being blank

gahtan-syarif:
is this code correct? not sure

### DIFF
--- a/src/matchmaking/tournament/tournament_manager.cpp
+++ b/src/matchmaking/tournament/tournament_manager.cpp
@@ -77,7 +77,7 @@ void TournamentManager::validateEngines() const {
     }
 
     for (std::size_t i = 0; i < engine_configs_.size(); i++) {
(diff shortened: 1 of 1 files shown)


src/matchmaking/tournament/tournament_manager.cpp
```diff
@@ -77,6 +77,7 @@ void TournamentManager::validateEngines() const {
     }
 
     for (std::size_t i = 0; i < engine_configs_.size(); i++) {
+        if (engine_configs_[i].name == "") {throw std::runtime_error("Error; please specify a name for each engine!");}
```
Disservin, line 80: sure is correct, the more c++ idomatic way would be
```cpp
if (engine_configs_[i].name.empty())
```

gahtan-syarif: fixed 👍

PR #349, opened 2024-05-03
feat: support more cutechess style inputs

gahtan-syarif:
cutechess users sometimes specify tc=10+0.1 as tc=inf/10+0.1 , this patch now supports that.

also adds support for cutechess style epdout where they use -epdout NAME instead of -epdout file=NAME

already tested it, works 👍 can safely merge


gahtan-syarif:
the only thing missing ig is sometimes cutechess users specify -repeat 2 instead of just -repeat, dont know how to do that one tho

PR #352, opened 2024-05-04
feat: accept -repeat 2 input

gahtan-syarif:
a lot of times cutechess users would do -repeat 2 instead of -repeat, this patch now treats both of them as the same instead of throwing a parsing error

already tested it, works 👍


gahtan-syarif:
hmm how do i check if argv[i] is a number?

gahtan-syarif:
> ```c++
> bool is_number(const std::string &s) {
>     return !s.empty() && std::find_if(s.begin(), s.end(), [](unsigned char c) { return !std::isdigit(c); }) == s.end();
> }
> ```
> 
> `is_number(argv[i])`
hmm this doesnt seem to work if i just specify -repeat without any additional inputs
```
void parseRepeat(int &i, int argc, char const *argv[], ArgumentData &argument_data) {
    try {
        parseValue(i, argc, argv, argument_data.tournament_options.games);
        if(!is_number(argv[i])) throw std::runtime_error("");
    } catch (const std::exception &e) {
        argument_data.tournament_options.games = 2;
    }
}
```

Disservin:
Ah might have to be i +1. what is argv of i there?

gahtan-syarif:
> Ah might have to be i +1. what is argv of i there?

ah yes, argv[i] seems to be "-repeat"

gahtan-syarif:
fixed 👍

PR #353, opened 2024-05-05
feat: support minutes:seconds format in tc

gahtan-syarif:
in cutechess you can specify minutes in the tc
```
     Play a single game between Atak and Glaurung engines with a time control
     of 40 moves in one minute and 30 seconds with one second increment:

	   $ cutechess-cli -engine name=Atak cmd=Atak32.exe dir=C:\atak
	   proto=xboard -engine cmd=glaurung proto=uci option.Threads=1 -each
	   tc=40/1:30+1
```
with this patch fastchess can now do that too

tested it, works 👍


src/cli/cli.cpp
```diff
@@ -76,12 +77,19 @@ TimeControl::Limits parseTc(const std::string &tcString) {
...
         remainingStringVector = inc[0];
     }
-
-    tc.time = static_cast<int64_t>(std::stod(remainingStringVector) * 1000);
+  
+    if (has_minutes) {
+        const auto minutes_string = str_utils::splitString(remainingStringVector, ':');
+        int64_t minutes = static_cast<int64_t>(std::stod(minutes_string[0]) * 1000);
+        int64_t seconds = static_cast<int64_t>(std::stod(minutes_string[1]) * 1000);
+        tc.time = minutes * 60 + seconds;
```
Disservin, line 89: the names are confusing btw, because minutes isn't really minutes anymore

Disservin:
wasnt entirely sure if i'd want this but i guess it will be nice, needs a documentation change too

src/cli/cli.cpp
gahtan-syarif: ahh yes, i should have multiplied it by 1000 after the variable is initialized

gahtan-syarif:
> wasnt entirely sure if i'd want this but i guess it will be nice, needs a documentation change too

ill change the man in a followup pr alongside other fixes

also i renamed the variable for less confusion

src/cli/cli.cpp
gahtan-syarif: fixed 👍

PR #356, opened 2024-05-08
chore: update man

gahtan-syarif:
also simplified autosaveinterval code


src/matchmaking/tournament/roundrobin/roundrobin.cpp
```diff
@@ -22,10 +22,9 @@ RoundRobin::RoundRobin(const options::Tournament& tournament_config,
 void RoundRobin::start() {
     BaseTournament::start();
     // Check if the user wants to automatically save the results
-    const auto do_autosave = tournament_options_.autosave.enabled;
-    if (do_autosave){
+    if (tournament_options_.autosaveinterval != 0){
```
Disservin, line 25: make the interval unsigned or say > 0

Disservin (changes requested):

src/matchmaking/tournament/roundrobin/roundrobin.cpp
gahtan-syarif: fixed 👍 @Disservin

Issue #335, opened 2024-05-01
different behaviour of draw and resign adjudication compared to cutechess


Disservin:
vondele saw different results with 
``` 
-draw movenumber=50 movecount=8 score=5 -resign movecount=3 score=600
```
compared to what cutechess outputs

gahtan-syarif:
> vondele saw different results with
> 
> ```
> -draw movenumber=50 movecount=8 score=5 -resign movecount=3 score=600
> ```
> 
> compared to what cutechess outputs

for resignation this is explainable, cutechess has one-sided resignation by default while fastchess only offers two-sided resignation. from cutechess docs
```
     -resign movecount=count score=score [twosided=value]
	     Adjudicate the game as a loss if an engine's score is at least
	     score centipawns below zero for at least count consecutive moves.
	     If value is true (default: false) then activate two-sided resign
	     adjudication.  The winning side's scores must be at least score
	     centipawns above zero for at least count consecutive moves
```

gahtan-syarif:
another reason is resign and draw adjudication in cutechess starts its counter from every odd plies, while in fastchess the counter starts from both odd and even plies @Disservin

gahtan-syarif:
> another reason is resign and draw adjudication in cutechess starts its counter from every odd plies, while in fastchess the counter starts from both odd and even plies @Disservin

for this i think this one might be considered a bug, since if the counter starts counting from an even ply it would start comparing one engine's eval for this move with the eval of the opponent from the previous move

gahtan-syarif:
another reason for the difference: captures and pawn moves reset the adjudicate draw counter in cutechess while in fastchess it doesnt
```
     -draw movenumber=number movecount=count score=score
	     Adjudicate the game as draw if the score of both engines is
	     within score centipawns from zero for at least count consecutive
	     moves, and at least number full moves have been played.  Captures
	     and pawn moves will reset the counters.
```

PR #357, opened 2024-05-08
fix: draw and resign counter now starts from odd plies only

gahtan-syarif:
prevents eval from being compared to eval of previous moves. tested it and seems to work ok.

partly solves https://github.com/Disservin/fast-chess/issues/335


gahtan-syarif:
now the only thing left to make it 1 to 1 with cutechess adjudication is to have the draw counter reset after ever capture/pawn move and to have one-sided resignation available.

PR #359, opened 2024-05-08
feat: make fastchess draw adjudication equivalent to cutechess

gahtan-syarif:
this patch makes fast-chess adjudication equivalent to cutechess, comparison can be seen below

prompt: 
```
fast-chess -draw movenumber=3 movecount=3 score=15 -engine cmd=berserk-13 name=berserk -engine cmd=berserk-13 name=berserk2 -each tc=inf nodes=1000000 proto=uci -pgnout adjudication-test-fastchess -rounds 1 -games 1

cutechess-cli -draw movenumber=3 movecount=3 score=15 -engine cmd=berserk-13 name=berserk -engine cmd=berserk-13 name=berserk2 -each tc=inf nodes=1000000 proto=uci -pgnout adjudication-test-cutechess -rounds 1 -games 1
```

resulting pgn:
[adjudication-test-fastchess.txt](https://github.com/Disservin/fast-chess/files/15255463/adjudication-test-fastchess.txt)
[adjudication-test-cutechess.txt](https://github.com/Disservin/fast-chess/files/15255465/adjudication-test-cutechess.txt)

as can be seen, both are equivalent.

partially solves issue https://github.com/Disservin/fast-chess/issues/335 (one-sided resignation havent been added, will work on it soon).

### DIFF
--- a/src/engine/process/process_posix.hpp
+++ b/src/engine/process/process_posix.hpp
@@ -33,10 +33,8 @@
 
 namespace fast_chess {
 extern util::ThreadVector<pid_t> process_list;
-}
 
-namespace fast_chess::engine::process {
-
+namespace engine::process {
 class Pipes {
     int pipe_[2];
 
@@ -137,7 +135,7 @@ class Process : public IProcess {
             process_pid_ = fork_pid;
 
             // append the process to the list of running processes
-            fast_chess::process_list.push(process_pid_);
+            process_list.push(process_pid_);
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/matchmaking/match/match.hpp
```diff
@@ -17,10 +17,16 @@ class DrawTracker {
         draw_score   = tournament_config.draw.score;
     }
 
-    void update(const int score, const int move_count, engine::ScoreType score_type) noexcept {
+    void update(const int score, const int move_count, engine::ScoreType score_type, chess::Color color,
+               std::string Fen) noexcept {
```
Disservin, line 21: why get the fen only to split the fen and parse it? call the according functions on the board object to get the hmvc

Disservin (changes requested):

src/matchmaking/match/match.hpp
gahtan-syarif: > why get the fen only to split the fen and parse it? call the according functions on the board object to get the hmvc

i see, whats the function to get the hmvc from the board object?

Disservin: well im using my chess lib which got it's docs here https://disservin.github.io/chess-library/pages/usage.html

Issue #354, opened 2024-05-06
inaccurate pgn comments when multipv>1


gahtan-syarif:
the pgn comments containing the nodes nps seldepth and centipawn evaluation does not display the main PV info when multipv is set to above 1. you can see in the pgn and log file below where the pgn got its info from the multipv2 info string instead of the multipv1 info string
[log.txt](https://github.com/Disservin/fast-chess/files/15218719/log.txt)
[pgn.txt](https://github.com/Disservin/fast-chess/files/15218720/pgn.txt)
ideally when parsing the info strings we only take information from the multipv1 info string

Disservin:
fixed by https://github.com/Disservin/fast-chess/pull/360

PR #360, opened 2024-05-09
fix: only parse multipv 1 info string

gahtan-syarif:
fix issue https://github.com/Disservin/fast-chess/issues/354

havent tested it, but i think it works :P

### DIFF
--- a/src/engine/uci_engine.cpp
+++ b/src/engine/uci_engine.cpp
@@ -98,10 +98,17 @@ std::string UciEngine::lastInfoLine() const {
     // iterate backwards over the output and save the first line
     // that contains "info depth" and score and multipv 1
     for (auto it = output_.rbegin(); it != output_.rend(); ++it) {
-        if (it->line.find("info") != std::string::npos &&
-            it->line.find(" score ") != std::string::npos &&
-            it->line.find(" multipv 1 ") != std::string::npos) {
-            return it->line;
+        if (it->line.find(" multipv 1 ") != std::string::npos){
+            if (it->line.find("info") != std::string::npos &&
+                it->line.find(" score ") != std::string::npos &&
+                it->line.find(" multipv 1 ") != std::string::npos) {
+                return it->line;
+            }
+        } else {
+            if (it->line.find("info") != std::string::npos &&
+                it->line.find(" score ") != std::string::npos) {
+                return it->line;
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


Disservin:
this wont work, people who dont have multipv wont report multipv, if multipv exists in the output then it needs to be 1, if it doesnt then old logic

gahtan-syarif:
> this wont work, people who dont have multipv wont report multipv, if multipv exists in the output then it needs to be 1, if it doesnt then old logic

something like this then perhaps?
```
std::string UciEngine::lastInfoLine() const {
    // iterate backwards over the output and save the first line
    // that contains "info depth" and score and multipv 1
    for (auto it = output_.rbegin(); it != output_.rend(); ++it) {
        if (it->line.find(" multipv 1 ") != std::string::npos){
            if (it->line.find("info") != std::string::npos &&
                it->line.find(" score ") != std::string::npos &&
                it->line.find(" multipv 1 ") != std::string::npos) {
                return it->line;
            }
        } else {
            if (it->line.find("info") != std::string::npos &&
                it->line.find(" score ") != std::string::npos) {
                return it->line;
            }
        }
    }

    return {};
}

```

Disservin (commented):
I think this can be just
```cpp
    for (auto it = output_.rbegin(); it != output_.rend(); ++it) {
        if (it->line.find("info") != std::string::npos &&
            it->line.find(" score ") != std::string::npos &&
            (it->line.find(" multipv ") == std::string::npos ||
             it->line.find(" multipv 1") != std::string::npos)) {
            return it->line;
        }
    }
```
?

gahtan-syarif:
> I think this can be just
> 
> ```c++
>     for (auto it = output_.rbegin(); it != output_.rend(); ++it) {
>         if (it->line.find("info") != std::string::npos &&
>             it->line.find(" score ") != std::string::npos &&
>             (it->line.find(" multipv ") == std::string::npos ||
>              it->line.find(" multipv 1") != std::string::npos)) {
>             return it->line;
>         }
>     }
> ```
> 
> ?

wouldnt this return the first multipv it finds? even if its multipv 2?

gahtan-syarif:
ahh wait, yeah i didnt see the ==

gahtan-syarif:
updated

src/engine/uci_engine.cpp
```diff
@@ -96,10 +96,12 @@ void UciEngine::writeLog() const {
 
 std::string UciEngine::lastInfoLine() const {
     // iterate backwards over the output and save the first line
-    // that contains "info depth" and score
+    // that contains "info depth" and score and multipv 1
```
Disservin, line 99: well this is wrong? it doesnt have to contain multipv 1..

gahtan-syarif: fixed 👍

PR #371, opened 2024-05-09
fix makefile

gahtan-syarif:
findstring doesnt seem to work

this patch doesnt fix for clang but works for mingw


Disservin:
a) this way isn't a good way to do it
b) obviously the error is in that `VARIABLE` is wrong and it should be CXX..

Disservin:
https://github.com/Disservin/fast-chess/commit/ef185f4443a25e802ef9064cd66a927e10108fab

PR #375, opened 2024-05-10
fix: exact ep square handling

Disservin:
fixes the issue where
```
[Variant "From Position"]
[FEN "6k1/1p2p1rp/rP1pR3/2pP1pP1/p1P2P1P/R5K1/8/8 b - - 0 1"]

1... h5 2. Rh6 Rh7 3. Re6 Rg7 4. Rh6 Rh7 5. Re6 Rg7
```

is correctly adjudicated as a threefold repetition.

### DIFF
--- a/app/src/game/book/pgn_reader.cpp
+++ b/app/src/game/book/pgn_reader.cpp
@@ -52,7 +52,7 @@ class PGNVisitor : public chess::pgn::Visitor {
             return;
         }
 
-        board_.makeMove(move_i);
+        board_.makeMove<true>(move_i);
 
         if (board_.isGameOver().second != chess::GameResult::NONE) {
             early_stop_ = true;
--- a/app/src/game/epd/epd_builder.hpp
+++ b/app/src/game/epd/epd_builder.hpp
@@ -23,7 +23,7 @@ class EpdBuilder {
 
             if (illegal) break;
 
-            board.makeMove(chess::uci::uciToMove(board, move.move));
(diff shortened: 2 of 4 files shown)


gahtan-syarif:
is this pr not ready?

gahtan-syarif:
It seems cutechess has went ahead and implemented this https://github.com/cutechess/cutechess/commit/b57c70259c5dfedd94e3fd72747abf0b4fd77fd2

Disservin:
@ArcheCadia, @gahtan-syarif I am still a bit unsure about this..
while we check the engines pv we also check if the game is over 
https://github.com/Disservin/fastchess/blob/680feb0b4b074d6f48995ef139180bcba97962e6/app/src/matchmaking/match/match.cpp#L497
also taking the 3fold repetition into account https://github.com/Disservin/fastchess/blob/680feb0b4b074d6f48995ef139180bcba97962e6/app/src/matchmaking/match/match.cpp#L40 
with the change fastchess will notice (correctly) that the game already ended, most engines do not implement this correctly and as such will get illegal pv warnings after this merge.. I guess it is technically correct, just thinking about the consequences of this.. does Stockfish do this even correctly or should we skip the repetition check for the pv line?

gahtan-syarif:
Personally i think its a good idea to let engine authors know through the illegal pv warning. Its an edge case thats not taken into account in a lot of engines because they are not aware... so by doing this it will give them a heads up to implement correct ep square handling on pinned positions

PR #378, opened 2024-05-11
refactor: seperate score result from rating report for cutechess output

gahtan-syarif:
mimic cutechess behavior and also makes it more easily parseable i think


src/matchmaking/output/output_cutechess.hpp
```diff
@@ -35,6 +35,16 @@ class Cutechess : public IOutput {
...
 
+        std::cout << ss.str() << std::flush;
+    }
+
+    void printElo(const Stats& stats, const std::string& first,
+                  const std::string& second) override {
+        const elo::EloWDL elo(stats);
+
+         //get rid of unused parameter compiler warning
+        if (false) std::cout << first << second;
```
Disservin, line 46: Remove the name from the parameter list

src/matchmaking/output/output_fastchess.hpp
```diff
@@ -19,6 +19,13 @@ class Fastchess : public IOutput {
         std::cout << "--------------------------------------------------\n";
     };
 
+    void printResult(const Stats& stats, const std::string& first,
+                     const std::string& second) override {
+        //do nothing
+        //get rid of unused parameter compiler warning
+        if (false) std::cout << first << second << stats.wins;
```
Disservin, line 26: Same

Disservin (changes requested):

src/matchmaking/output/output_cutechess.hpp
gahtan-syarif: wouldnt it just error?

Disservin:
```
(const Stats&, const std::string&, const std::string&)
```

gahtan-syarif: updated

src/matchmaking/output/output_fastchess.hpp
gahtan-syarif: updated

PR #382, opened 2024-05-12
feat: make resignation same as cutechess resignation

gahtan-syarif:
consistent with cutechess

### DIFF
--- a/src/matchmaking/match/match.cpp
+++ b/src/matchmaking/match/match.cpp
@@ -234,7 +234,7 @@ bool Match::playMove(Player& us, Player& opponent) {
         return false;
     }
 
-    resign_tracker_.update(us.engine.lastScore(), us.engine.lastScoreType());
+    resign_tracker_.update(us.engine.lastScore(), us.engine.lastScoreType(), board_.sideToMove());
     maxmoves_tracker_.update();
 
     board_.makeMove(move);
--- a/src/matchmaking/match/match.cpp
+++ b/src/matchmaking/match/match.cpp
@@ -340,7 +340,7 @@ bool Match::adjudicate(Player& us, Player& them) noexcept {
         data_.termination = MatchTermination::ADJUDICATION;
         data_.reason      = us.engine.getConfig().name;
 
-        if (us.engine.lastScore() < tournament_options_.resign.score) {
+        if (us.engine.lastScore() < 0) {
             setLose(us, them);
             data_.reason += Match::ADJUDICATION_LOSE_MSG;
         } else {
--- a/src/matchmaking/match/match.cpp
+++ b/src/matchmaking/match/match.cpp
@@ -234,13 +234,12 @@ bool Match::playMove(Player& us, Player& opponent) {
         return false;
     }
 
-    resign_tracker_.update(us.engine.lastScore(), us.engine.lastScoreType(), board_.sideToMove());
-    maxmoves_tracker_.update();
-
     board_.makeMove(move);
 
     draw_tracker_.update(us.engine.lastScore(), board_.fullMoveNumber(), us.engine.lastScoreType(), 
                          board_.halfMoveClock());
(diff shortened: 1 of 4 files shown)


gahtan-syarif:
@Disservin how do i differentiate a negative mate score from a positive mate score? https://github.com/gahtan-syarif/fast-chess/blob/onesidedresign/src/matchmaking/match/match.hpp#L54

gahtan-syarif:
because rn it adjudicates as a loss even though its a positive mate `79. Kb1 {+M9/39, 0.047s, Engine1 loses by adjudication} 0-1`

gahtan-syarif:
ahh the score is the moves till mate

src/matchmaking/match/match.hpp
```diff
@@ -46,27 +46,45 @@ class ResignTracker {
...
-    void update(const int score, engine::ScoreType score_type) noexcept {
+    void update(const int score, engine::ScoreType score_type, chess::Color color) noexcept {
         if ((std::abs(score) >= resign_score && score_type == engine::ScoreType::CP) ||
             score_type == engine::ScoreType::MATE) {
-            resign_moves++;
+            if (twosided_ && (color == chess::Color::WHITE || resign_moves > 0)) resign_moves++;
         } else {
             resign_moves = 0;
         }
+        if (color == chess::Color::BLACK && !twosided_) {
```
Disservin, line 59: simplify this?
```cpp
if (!twosided_) {
    int& resign_moves = (color == chess::Color::BLACK) ? resign_moves_black : resign_moves_white;

    if ((score <= -resign_score && score_type == engine::ScoreType::CP) || 
        (score < 0 && score_type == engine::ScoreType::MATE)) {
        resign_moves++;
    } else {
        resign_moves = 0;
    }
}
```

src/matchmaking/match/match.hpp
```diff
@@ -46,27 +46,45 @@ class ResignTracker {
...
+               (score < 0 && score_type == engine::ScoreType::MATE)) {
+                resign_moves_white++;
+           } else {resign_moves_white = 0;}
+       }
     }
 
-    [[nodiscard]] bool resignable() const noexcept { return resign_moves >= move_count_ * 2; }
+    [[nodiscard]] bool resignable() const noexcept { 
+       if (twosided_) return resign_moves >= move_count_ * 2;
+       else return (resign_moves_black >= move_count_ || resign_moves_white >= move_count_);
```
Disservin, line 74: why else

src/matchmaking/match/match.cpp
```diff
@@ -340,7 +339,7 @@ bool Match::adjudicate(Player& us, Player& them) noexcept {
         data_.termination = MatchTermination::ADJUDICATION;
         data_.reason      = us.engine.getConfig().name;
 
-        if (us.engine.lastScore() < tournament_options_.resign.score) {
+        if (us.engine.lastScore() < 0) {
```
Disservin, line 342: resign score is now useless?

gahtan-syarif: yeah its not needed there, its better this way actually, since there will no longer be cases where the engine with positive mate score get wrongly adjudicated as a loss

src/matchmaking/match/match.hpp
gahtan-syarif: fixed 👍

src/matchmaking/match/match.hpp
gahtan-syarif: fixed

src/matchmaking/match/match.cpp
Disservin: then we can remove the option?

gahtan-syarif: > then we can remove the option?

no i meant its not needed in match.cpp, the one in match.hpp is still necessary

Disservin: ah right thanks

PR #386, opened 2024-05-13
refactor: sleep in busy loop

Disservin:


gahtan-syarif:
```c++
std::this_thread::sleep_for(std::chrono::seconds(1));
```
what does this line do?

Disservin:
> ```c++
> std::this_thread::sleep_for(std::chrono::seconds(1));
> ```
> 
> what does this line do?
> 
> edit: apparently its supposed to add a delay between games? if so can we make this as an option instead and set the default as 0 like cutechess? https://github.com/cutechess/cutechess/blob/master/docs/cutechess-cli.6.txt#L242

no this is the mainthread which waits for the tournament to end

Disservin:
what i dont get is why the ci is bugged out

gahtan-syarif:
> > ```c++
> > std::this_thread::sleep_for(std::chrono::seconds(1));
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > what does this line do?
> > edit: apparently its supposed to add a delay between games? if so can we make this as an option instead and set the default as 0 like cutechess? https://github.com/cutechess/cutechess/blob/master/docs/cutechess-cli.6.txt#L242
> 
> no this is the mainthread which waits for the tournament to end

ahh ok

gahtan-syarif:
> what i dont get is why the ci is bugged out

is it for this pr specifically?

Disservin:
no and my last two commits on master

gahtan-syarif:
> no and my last two commits on master

can you try merging this branch to latest master?

gahtan-syarif:
works

PR #399, opened 2024-05-16
feat: make pgnout min option available for cutechess pgnout

gahtan-syarif:
make fastchess more compatible with cutechess commands

tested it, works 👍


gahtan-syarif:
@Disservin how about this one should we add it or not? what it does is basically add support for `-pgnout FILE` min that cutechess users sometimes use, when its specified it enables minimal pgn format

PR #404, opened 2024-05-16
feat: make release build clang compatible

gahtan-syarif:
--no-as-needed flag doesnt seem to work with clang

when running `make -j build=release CXX=clang++`
```bash
pgn/pgn_builder.o tmp/src/pgn/pgn_reader.o tmp/src/time/timecontrol.o tmp/src/util/logger/logger.o -lpthread -static -static-libgcc -static-libstdc++ -Wl,--no-as-needed
lld: error: unknown argument: --no-as-needed
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [Makefile:121: fast-chess] Error 1
```
--no-as-needed flag also doesnt seem to do anything with GCC, it produces the same binary size regardless of whether or not it is specified, so might as well remove it


Disservin:
idk see the point of this you are basically going back to the discussion we had a few days ago regarding this entire thing.. and currently the binaries from the ci work and that is all what we need no need to get clang there working.. and if anyone on fishtest/openbench wants to build their own executable then they should just use `make -j` anyway

gahtan-syarif:
> idk see the point of this you are basically going back to the discussion we had a few days ago regarding this entire thing.. and currently the binaries from the ci work and that is all what we need no need to get clang there working.. and if anyone on fishtest/openbench wants to build their own executable then they should just use `make -j` anyway

i mean i get your point but in this case that flag is just dead code, it produces the same exact binary compiling with it compared to compiling without it. clang support aside that flag can probably be removed just to reduce code length alone

gahtan-syarif:
i have no strong feelings for this pr and i just wanted this added so i can do clang static compile with build=release on my computer so if you dont merge it its fine for me but just saying it be a lot cooler if you did

Disservin:
> > idk see the point of this you are basically going back to the discussion we had a few days ago regarding this entire thing.. and currently the binaries from the ci work and that is all what we need no need to get clang there working.. and if anyone on fishtest/openbench wants to build their own executable then they should just use `make -j` anyway
> 
> i mean i get your point but in this case that flag is just dead code, it produces the same exact binary compiling with it compared to compiling without it. clang support aside that flag can probably be removed just to reduce code length alone

Well for your dev environment.. have you tested windows, mac, linux? Maybe it does t do something no but might prevent issues in the future?

Disservin:
Also when openbench or fishtest use it they compile it without static linking and without downloading a release a build so for that it doesn’t matter and executables are only generated from gcc anyway

gahtan-syarif:
> > > idk see the point of this you are basically going back to the discussion we had a few days ago regarding this entire thing.. and currently the binaries from the ci work and that is all what we need no need to get clang there working.. and if anyone on fishtest/openbench wants to build their own executable then they should just use `make -j` anyway
> > 
> > 
> > i mean i get your point but in this case that flag is just dead code, it produces the same exact binary compiling with it compared to compiling without it. clang support aside that flag can probably be removed just to reduce code length alone
> 
> Well for your dev environment.. have you tested windows, mac, linux? Maybe it does t do something no but might prevent issues in the future?

i highly doubt it, cuz the flag isnt even written correctly here, its non functional
Screenshot 2024-05-19 030218
Screenshot 2024-05-19 030432
this whole time it does nothing except disable clang compilation

Disservin:
im pretty sure it was required for correct clang compilation on mac, stockfish also has this..

gahtan-syarif:
> im pretty sure it was required for correct static compilation, stockfish also has this..

then it should be placed before the libraries specified, since so far the artifacts have been working fine without it then its not necessary
Screenshot 2024-05-19 031957

Disservin:
ok have a try at latest master and check if that works for you

gahtan-syarif:
> ok have a try at latest master and check if that works for you

different sort of error now
Screenshot 2024-05-19 034221
this was the flag from your recent commit i think, maybe we can make that flag conditionally set to only gcc compiler too?

PR #405, opened 2024-05-16
refactor: improve version name

gahtan-syarif:
same as stockfish


Disservin:
i dont see a reason to do this

gahtan-syarif:
> i dont see a reason to do this

it makes it less confusing, cuz the current version printed is 0.8.2 even though thats the last release, and you get inconsistency with the version output of the git version and the local version where the git version prints the full year while the local version prints the last two digits of the year, plus it just looks nicer this way

Disservin:
i prefer to keep it the way it is

gahtan-syarif:
> i prefer to keep it the way it is

but then how do we differentiate the stable releases from dev versions? cuz they both print alpha-0.8.2

PR #406, opened 2024-05-17
fix: make game end with losing player last move when twosided resign

gahtan-syarif:
for twosided resign only

handle cases where game is adjudicated one ply later compared to cutechess when black is the one winning


gahtan-syarif:
also got rid of the " loses by adjudication" message since that does not exist in cutechess

PR #408, opened 2024-05-17
chore: update man

gahtan-syarif:
add the new feature of tracking hashfull and tbhits, also remove the version since it has an outdated version number and can be done using -version anyway


man
```diff
@@ -233,6 +235,3 @@ REPORTING BUGS
 
 COPYRIGHT
         This software is licensed under the MIT license. See the LICENSE file for details.
-
-VERSION
```
Disservin, line 237: why

gahtan-syarif: the version is outdated, its written 0.8.2 there so i deleted it to remove confusion from the stable release version, plus i checked cutechess docs and they also dont have version

Disservin: why is 0.8.2 outdated? the first line also has the version.. `FAST-CHESS (alpha-0.8.2)            General Commands Manual             FAST-CHESS (alpha-0.8.2)`

gahtan-syarif: ahh should probably delete that too, change it to just fast-chess, this is how its written in cutechess docs
```
CUTECHESS-CLI(6)		 Games Manual		      CUTECHESS-CLI(6)
```

Disservin: i dont see the need to have the exact git version there as well

gahtan-syarif: the 0.8.2 is the last release right, this is the dev version for 0.8.3/0.9

PR #411, opened 2024-05-18
fix: allow policy=round command for openings

gahtan-syarif:
apparently policy=default and policy=round does pretty much the same thing when the number of engines is 2


Disservin:
mh im starting to think that backporting the cutechess options isnt good, this will make people think that fastchess supports it.. im fine with having an alternative syntax for stuff which is supported but not supported stuff should just throw an error

gahtan-syarif:
> mh im starting to think that backporting the cutechess options isnt good, this will make people think that fastchess supports it.. im fine with having an alternative syntax for stuff which is supported but not supported stuff should just throw an error

technically fastchess has this feature by default, since we switch openings every round, what we dont have is policy=encounter where we switch openings every new player (cuz we only do h2h matches)

PR #418, opened 2024-05-19
refactor: improve compatibility with supported cutechess command

gahtan-syarif:
adds support for `-pgnout FILE min` and allows `policy=round` input which fastchess has support for

### DIFF
--- a/src/cli/cli.cpp
+++ b/src/cli/cli.cpp
@@ -266,7 +266,14 @@ void parsePgnOut(const std::vector<std::string> &params, ArgumentData &argument_
         });
     } catch (const std::exception &e) {
         // try to read as cutechess pgnout
-        parseValue(params, argument_data.tournament_options.pgn.file);
+        std::string val = readUntilDash(i, argc, argv);
+        if (str_utils::contains(val, " ")) {
+            const auto string_vector = str_utils::splitString(val, ' ');
+            argument_data.tournament_options.pgn.file = string_vector[0];
+            if (str_utils::contains(val, " min")) argument_data.tournament_options.pgn.min = true;
+        } else {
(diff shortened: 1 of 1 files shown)


src/cli/cli.cpp
```diff
@@ -266,7 +266,15 @@ void parsePgnOut(const std::vector<std::string> &params, ArgumentData &argument_
         });
     } catch (const std::exception &e) {
         // try to read as cutechess pgnout
-        parseValue(params, argument_data.tournament_options.pgn.file);
+        std::string val = concat(params);
+        if (str_utils::contains(val, " ")) {
+            const auto string_vector = str_utils::splitString(val, ' ');
```
Disservin, line 271: you concat the params to just split the string again?

gahtan-syarif: ahh yeah that was from before the refactor, fixed it

src/cli/cli.cpp
```diff
@@ -266,7 +266,9 @@ void parsePgnOut(const std::vector<std::string> &params, ArgumentData &argument_
         });
     } catch (const std::exception &e) {
         // try to read as cutechess pgnout
-        parseValue(params, argument_data.tournament_options.pgn.file);
+        argument_data.tournament_options.pgn.file = params[0];
+        if (std::find(params.begin(), params.end(), "min") != params.end())
+            argument_data.tournament_options.pgn.min = true;
```
Disservin, line 271:
```
argument_data.tournament_options.pgn.min = std::find(params.begin(), params.end(), "min") != params.end();
```

gahtan-syarif: fixed 👍

PR #422, opened 2024-05-19
chore: update man

gahtan-syarif:


gahtan-syarif:
> ??

i think you wrote it the opposite way, stats=true keeps the stats while stats=false drops it

Disservin:
no https://github.com/Disservin/fast-chess/blob/master/src/cli/cli.cpp#L474

gahtan-syarif:
> no https://github.com/Disservin/fast-chess/blob/master/src/cli/cli.cpp#L474

yes, if you type in stats=false it drops the stats like the code you linked, while stats=true keeps it, and the default is it keeps it ive tested it

Disservin:
oh sorry yea youre right

Issue #424, opened 2024-05-21
something wrong with the dir command?


gahtan-syarif:
```
C:\Users\Gahtan Syarif Nahdi\Downloads\fast-chess-windows-latest>fast-chess.exe -openings file=C:/cutechess-1.3.1-win64/books/UHO_Lichess_4852_v1.epd format=epd -engine cmd=stockfish_16.1_x64.exe name=Engine1 -engine cmd=stockfish_16.1_x64.exe name=Engine2 -each tc=10+0.1 dir="C:\Chess Engines" -rounds 4 -concurrency 4 -resign movecount=1 score=600 -pgnout file=fastchess.pgn -ratinginterval 1 -log file=log -randomseed -pgnout C:/cutechess-1.3.1-win64/books/justatest.pgn
Starting tournament...
Started game 1 of 8 (Engine2 vs Engine1)
Started game 4 of 8 (Engine2 vs Engine1)
Started game 3 of 8 (Engine1 vs Engine2)
Started game 2 of 8 (Engine1 vs Engine2)
Exception RoundRobin:::playGame Warning; Something went wrong when pinging the engine.
Exception RoundRobin:::playGame Warning; Something went wrong when pinging the engine.
Exception RoundRobin:::playGame Warning; Something went wrong when pinging the engine.
Exception RoundRobin:::playGame Warning; Something went wrong when pinging the engine.
Finished match
Saved results.
```
doesnt happen when i dont use dir and instead specify the path directly in the cmd

gahtan-syarif:
i think i figured it out, it seems dir="C:/Chess Engines/" seems to work, so the issue is two things:
1. didnt automatically add the extra "/" on the end of the dir
2. didnt convert "\" into "\\" with the escape character so it can be interpreted correctly

Disservin:
logic here needs to be changed from a string to a filesystem path https://github.com/Disservin/fast-chess/blob/master/src/engine/uci_engine.cpp#L46
https://en.cppreference.com/w/cpp/filesystem/path
also needs the same handling of the filesystem library as done here https://github.com/Disservin/fast-chess/blob/master/src/cli/cli.cpp#L203

PR #427, opened 2024-05-29
refactor: better path joining of cmd and dir options

Disservin:


Disservin:
might fix https://github.com/Disservin/fast-chess/issues/424

gahtan-syarif:
works for me on windows 👍

Issue #439, opened 2024-06-10
Windows: high memory usage even with low number of rounds


gahtan-syarif:
![Screenshot 2024-06-11 025013](https://github.com/Disservin/fast-chess/assets/155860115/6b1278ef-c0b1-4488-bdb4-6f07be1125da)

command used:
```
fast-chess.exe -openings file=C:\cutechess-1.3.1-win64\books\UHO_Lichess_4852_v1.epd format=epd -engine cmd="Stockfish dev-20240521-c14b6979.exe" name=Engine1 -engine cmd="Stockfish dev-20240521-c14b6979.exe" name=Engine2 -each proto=uci tc=10+0.1 dir="C:/Chess Engines/" -rounds 10 -concurrency 4 -resign movecount=3 score=600 -pgnout file=fastchess.pgn -ratinginterval 1 -log file=log -randomseed -sprt elo0=0 elo1=5 alpha=0.05 beta=0.05 model=normalized
```
interestingly, in some cases just leaving things alone for long enough the memory usage would suddenly drop down to like 60MB

gahtan-syarif:
> u sure this isnt just the memory for the stockfish engines?

each stockfish instance memory usage is listed seperately on task manager, also i increased the hash size to 32 and it didnt affect the memory usage for fast-chess, see below
![Screenshot 2024-06-11 031227](https://github.com/Disservin/fast-chess/assets/155860115/5931e213-3b9d-4e82-a265-8c8312af419d)

vondele:
The book itself is 167MB and is kept in memory, I assume. If for whatever reason a second copy of it is made, you're at 334MB

gahtan-syarif:
it seems vondele is correct, this is without book:
![Screenshot 2024-06-11 032927](https://github.com/Disservin/fast-chess/assets/155860115/736ca8b0-6cae-4fcf-9883-d5120597cdc5)

but in cutechess interestingly with the lichess book the memory usage is kept at 80MB

Disservin:
Mh the book shouldn’t be in memory twice I’ll have to check how that happens

Disservin:
I think two things are happening.
A) Windows task manager updates slowly the memory usage 
B) the code is a bit suboptimal, i see about ~270MB usage, for the same config, still a bit too much given that the read book should be around 167MB. Still some 100MB short which im currently unsure where that ends up being wasted.

```cpp
        std::ifstream openingFile;
        openingFile.open(file);

        std::string line;
[...]

PR #440, opened 2024-06-10
refactor: reduce memory usage when using large books

gahtan-syarif:
reduce memory usage when using books https://github.com/Disservin/fast-chess/issues/439

before:
Screenshot 2024-06-11 025013

after:
Screenshot 2024-06-11 055940

idk how cutechess does it, but they somehow manage to keep memory usage fixed at 80MB even with 1 million games, do they somehow not load the book into memory?

### DIFF
--- a/src/book/opening_book.cpp
+++ b/src/book/opening_book.cpp
@@ -62,8 +62,8 @@ void OpeningBook::setup(const std::string& file, FormatType type) {
              }
         }
         
-        if (epd_.size() > rounds_) {
-            epd_.erase(epd_.begin() + static_cast<std::size_t>(rounds_), epd_.end());
+        if (epd_.size() > static_cast<std::size_t>(rounds_)) {
+            epd_.erase(epd_.begin() + rounds_, epd_.end());
             epd_.shrink_to_fit();
         }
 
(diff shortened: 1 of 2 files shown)


Disservin:
Why are you removing the variant?

gahtan-syarif:
> Why are you removing the variant?

i feel its unecessary, and i suspect its the reason the book gets duplicated in memory

Disservin:
> > Why are you removing the variant?
> 
> i feel its unecessary, and i suspect its the reason the book gets duplicated in memory

No, the vectors contents are heap allocated and the variant will be large enough to hold the largest member which doesn’t include the vectors heap memory. 
It also makes enforces that only 1 book is used, not two.

gahtan-syarif:
> > > Why are you removing the variant?
> > 
> > 
> > i feel its unecessary, and i suspect its the reason the book gets duplicated in memory
> 
> No, the vectors contents are heap allocated and the variant will be large enough to hold the largest member which doesn’t include the vectors heap memory. It also makes enforces that only 1 book is used, not two.

i see, but other than that i dont really know how its possible for it to get duplicated as the epd and pgn variable are auto deallocated after the setup function is finished and the book_ variable is the only one containing the book

gahtan-syarif:
something is wrong, even after the vector size reduction im still getting the same amount of memory usage when number of rounds is 10000

gahtan-syarif:
added shrink to fit, should work now

gahtan-syarif:
hmm just realized this isnt applicable to fishtest, as the starting index might be larger than the number of rounds, ill have to find some other way

PR #441, opened 2024-06-11
refactor: memory improvements for book handling

Disservin:


gahtan-syarif:
suggestion, i think maybe you can find a way to cut off all the openings that wouldnt be accessible during the game? so lets say you set -rounds=10, then the rest of the book vector outside of those 10 positions can be erased as theyll never come into play. 

this is what i tried to do with my last pr but i only made it work when the starting offset and initial matchcount is the default values because of my weak programming skills, so maybe youd like to take a shot at it?

PR #442, opened 2024-06-12
refactor: truncate unused openings in book vector to reduce memory usage

gahtan-syarif:
rotates the book vector based on the starting offset and truncates the unused parts to reduce memory usage.
https://github.com/Disservin/fast-chess/issues/439


gahtan-syarif:
i got the memory usage to go down by 40MB with the lichess book, i checked the .size() and .capacity() of the books and as expected, it reduces both to the number of rounds. everything seems to work well. theres still 200MB in the task manager, i suspect this is just the memory allocator not immediately returning the memory to the system until it is needed

Disservin:
now this is simply reverting the old commit ? xd 
if you load a full book you still have the memory for the book itself + sizeof string

gahtan-syarif:
> now this is simply reverting the old commit ? xd if you load a full book you still have the memory for the book itself + sizeof string

the full book is only loaded when you run 2.6 million rounds, otherwise the book vector would be truncated and memory size reduced with shrink_to_fit(), i think this makes more sense especially in fishtest like scenario where each worker runs only a couple hundred games each test

gahtan-syarif:
Screenshot 2024-06-14 023144
this is the memory usage with 10 rounds

gahtan-syarif:
can this be tested out with massif? in my system i got pretty decent memory usage reduction compared to master

Disservin:
why would this even need a memory test? before merging anything like this i'd rather have a unit test which covers the expected order of openings which are supposed to come out + how will this behave when fc is restarted from a config with games played?

gahtan-syarif:
> why would this even need a memory test? before merging anything like this i'd rather have a unit test which covers the expected order of openings which are supposed to come out + how will this behave when fc is restarted from a config with games played?

ive tested it independently and the resulting pgn is the same as master, also the config with games played + starting offset thing is already taken into account with the rotate() function that rotates the book based on the offset_ variable here https://github.com/gahtan-syarif/fast-chess/blob/reducemem/src/book/opening_book.cpp#L17 https://github.com/gahtan-syarif/fast-chess/blob/reducemem/src/book/opening_book.hpp#L38
i assure you theres nothing to worry about

Disservin:
manual testing is good, but unit tests catch bugs one might miss and ensures future changes don't break things..

gahtan-syarif:
> manual testing is good, but unit tests catch bugs one might miss and ensures future changes don't break things..

alright ill try to work on a unit test for it, but for this one im 100% confident that theres no bugs, ive tested it with starting offset, no starting offset, rounds greater than book size, etc... in fact, the only bug is in master where the last line doesnt get parsed when the text file doesnt end in a newline

PR #451, opened 2024-06-16
refactor: truncate unused openings and add unit tests for validation

gahtan-syarif:
i added the unit tests as requested, pls merge 🙏

### DIFF
--- a/src/book/opening_book.cpp
+++ b/src/book/opening_book.cpp
@@ -15,6 +15,7 @@ OpeningBook::OpeningBook(const options::Tournament& tournament) {
     order_  = tournament.opening.order;
     plies_  = tournament.opening.plies;
     offset_ = start_ - 1 + matchcount_ / games_;
+    seed_   = tournament.seed;
     setup(tournament.opening.file, tournament.opening.format);
 }
 
--- a/src/book/opening_book.cpp
+++ b/src/book/opening_book.cpp
@@ -8,13 +8,13 @@
 
 namespace fast_chess::book {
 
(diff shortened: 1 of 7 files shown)


gahtan-syarif:
@Disservin help, idk how to access the openings.epd file in the tests, this doesnt seem to cut it: https://github.com/gahtan-syarif/fast-chess/blob/reducemem6/tests/openings_test.cpp#L12

gahtan-syarif:
> @Disservin help, idk how to access the openings.epd file in the tests, this doesnt seem to cut it: https://github.com/gahtan-syarif/fast-chess/blob/reducemem6/tests/openings_test.cpp#L12

nvm figured it out

Disservin:
This test isn't exactly testing much.. you are calling `getEpdBook` which returns a copy of the internal vector and then you check if certain strings are at the exact position you want them to be. You should be testing it how it is also used in the code, meaning you create the book and then you call `fetchId` multiple times and you use the result of that to index the `OpeningBook` which you then assert

gahtan-syarif:
> This test isn't exactly testing much.. you are calling `getEpdBook` which returns a copy of the internal vector and then you check if certain strings are at the exact position you want them to be. You should be testing it how it is used in the code, meaning you create the book and then you call `fetchId` multiple times and you use the result of that to index the `OpeningBook` which you then assert and you call fetch as often as the number of rounds which will be played. Also there are cases to cover,
> 
> 1. rounds < number of openings
> 2. rounds == number of openings
> 3. rounds > number of openings

alright, ill fix it

PR #455, opened 2024-06-17
refactor: enhance openings unit test and truncate unused openings

gahtan-syarif:
truncate the unused openings to save memory and incorporate the initial matchcount to the unit test. now all edge cases have been accounted for.

https://github.com/Disservin/fast-chess/issues/439


Disservin:
can we first get a fix for faster, i just saw that windows ci failed there? not sure why tbh

gahtan-syarif:
> can we first get a fix for faster, i just saw that windows ci failed there? not sure why tbh

the windows ci failed for your character array string view parsing thing, i think its because of the last line not parsed thing, this pr doesnt have that problem

Disservin:
well i fixed the last line parsing or so i had hoped... macos and ubuntu work, windows doesn't 
and i at least want this to be in master green before new stuff. I also don't understand why you changed the tests now? I'd expect them to look like they did before... changing the seed size should go in a separate pr

gahtan-syarif:
> well i fixed the last line parsing or so i had hoped... macos and ubuntu work, windows doesn't and i at least want this to be in master green before new stuff. I also don't understand why you changed the tests now? I'd expect them to look like they did before... changing the seed size should go in a separate pr

i tried your fix on my own private repo i think and it didnt work out very well, something to do with it parsing empty lines

as for the tests i incorporated initial matchcount directly cuz i thought it would be faster that way than having it be in a separate PR, thats why the fens also changed, for cases where initial matchcount=0 the fen should stay the same. i also moved the seed initialization to opening_book.hpp so the seed can be set from the test case directly, since before the seed initialization was taking place in the tournament manager.

and i changed the seed size to 64 bit since its just a minor change, we're already using the 64 bit mersenne twister anyway so that part just allows for setting greater seeds, it doesnt affect the shuffling

Disservin:
i still dont like this :/
a) for now a totally unnecessary change to the seed size -> should be it's own pr
b) hacky way to create the book now -> roundrobin creates it whereas the base tournament did this before
c) modifying the tests unnecessarily.. I realized `Test more openings than rounds` is actually doing it randomly and not sequential but so what for now. Truncating the book will change id's ok, but why do you set to the matchcount to 2 ? basically making it impossible to easily tell if this patch works with the current tests?

i will cleanup the stuff for b) a bit and fix the test case for c)

gahtan-syarif:
> i still dont like this :/ a) for now a totally unnecessary change to the seed size -> should be it's own pr b) hacky way to create the book now -> roundrobin creates it whereas the base tournament did this before c) modifying the tests unnecessarily.. I realized `Test more openings than rounds` is actually doing it randomly and not sequential but so what for now. Truncating the book will change id's ok, but why do you set to the matchcount to 2 ? basically making it impossible to easily tell if this patch works with the current tests?
> 
> i will cleanup the stuff for b) a bit and fix the test case for c)

for a) and c) i can just revert it, but for b) i moved it to roundrobin cuz thats the only way i can make it work with the initial matchcount being an input parameter, the void setinternaloffset() function doesnt work when the matchcount is used in the constructor

Disservin:
this should make b) work https://github.com/Disservin/fast-chess/commit/be0386f94b5ed807d6c38920e9a2aca101593fe8

gahtan-syarif:
alright, fixing the pr now

PR #456, opened 2024-06-17
fix logger thread id formatting

gahtan-syarif:
fix this:
Screenshot 2024-06-18 060804


Disservin:
there isn't really something wrong here, i think initially i tested this on windows which has these low id's where setw(3) makes sense.. it let's you easier distinguish between different id's because they have roughly the same length

gahtan-syarif:
> there isn't really something wrong here, i think initially i tested this on windows which has these low id's where setw(3) makes sense.. it let's you easier distinguish between different id's because they have roughly the same length... now on linux like systems i think the id is typically way larger

i see, closing this pr then

PR #460, opened 2024-06-18
refactor: change seed type to uint64_t

gahtan-syarif:
the std::mt19937_64 random number generator we use in rand.hpp is a 64 bit prng, and as such accepts 64 bit seeding. this allows for higher degree of randomness when using `-randomseed`


Disservin:
> this allows for higher degree of randomness

well this isn't really true but w/e

PR #461, opened 2024-06-19
refactor: better way of shrinking the book vector

gahtan-syarif:
the current way of shrinking the vector is by relying on the shrink_to_fit() function to get rid of excess memory usage. this is not ideal since shrink_to_fit() is implementation defined and a non-binding request, so its completely up to the memory allocator how much memory it frees up. with this patch the shrinking is done using the vector swap trick where the vector is swapped with a temporary copy that immediately goes out of scope. this 100% ensures complete memory reduction.

here is a comparison of the memory usage in a match of 1000 rounds using UHO lichess book with shrink_to_fit() method and vector swap method:

shrink_to_fit():
Screenshot 2024-06-19 155614

vector swap trick:
Screenshot 2024-06-19 155353

as can be seen, the new method is much more effective and reliable.

### DIFF
--- a/src/book/opening_book.cpp
+++ b/src/book/opening_book.cpp
@@ -57,7 +57,16 @@ void OpeningBook::setup(const std::string& file, FormatType type) {
 
     rotate(offset_);
     truncate(rounds_);
-    shrink();
+
+    if (type == FormatType::EPD) {
+        std::vector<std::string> tmp(std::get<epd_book>(book_).begin(), std::get<epd_book>(book_).end());
+        std::get<epd_book>(book_).swap(tmp);
+        std::get<epd_book>(book_).shrink_to_fit();
+    } else {
+        std::vector<pgn::Opening> tmp(std::get<pgn_book>(book_).begin(), std::get<pgn_book>(book_).end());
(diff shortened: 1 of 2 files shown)


src/book/opening_book.cpp
```diff
@@ -57,7 +57,17 @@ void OpeningBook::setup(const std::string& file, FormatType type) {
 
     rotate(offset_);
     truncate(rounds_);
-    shrink();
+
+    // force deallocation of memory
+    if (type == FormatType::EPD) {
```
Disservin, line 62: try this

```cpp
    std::visit([](auto& book) {
        using BookType = std::decay_t<decltype(book)>;
        std::vector<typename BookType::value_type> tmp(book.begin(), book.end());
        book.swap(tmp);
        book.shrink_to_fit();
    }, book);
```

Disservin (changes requested):

src/book/opening_book.cpp
gahtan-syarif: alright

gahtan-syarif: fixed 👍 works

PR #465, opened 2024-06-19
Proper rand

gahtan-syarif:
introduce proper rand functions that returns a random number between x and y. this is ideal since the current implementation just uses rand % N which introduces modulo bias https://stackoverflow.com/questions/10984974/why-do-people-say-there-is-modulo-bias-when-using-a-random-number-generator

changed the fens in opening_test.cpp to accomodate this upgrade


Disservin:
i dont really think there is a big need for this :p

gahtan-syarif:
yeah, it seems the uniform_int_distribution is not portable across platforms

PR #466, opened 2024-06-19
fix: make random device have 64 bit resolution

gahtan-syarif:
std::random_device returns a uint32_t, which means it can only generate 2^32 unique seeds. by calling random device twice and bitshifting and combining it, we can make a 64 bit random device suitable for seeding the 64 bit mersenne twister engine.

this allows `-randomseed` to return 2^64 different values instead of 2^32


Disservin:
why are you so obsessed with the random part, we don't need any advanced stuffed.. having 32 bit prng is more than enough

gahtan-syarif:
> why are you so obsessed with the random part, we don't need any advanced stuffed.. having 32 bit prng is more than enough

yeah but why not 64 bit while we're at it, and -randomseed right now only produces 2^32 unique values even though the output range is 2^64 so this fixes it to its intended behavior, also doesnt really add any complexity

PR #474, opened 2024-06-20
refactor: make uci engine communication more descriptive

Disservin:


Disservin:
@gahtan-syarif i tried to make some warnings and communication descriptive so that these are a bit more self explanatory when people encounter a warning, or when their engine crashes suddenly in which case the log will have some more information about the entire communication with the engine. Now with `trace` the log will grow rather fast.

PR #475, opened 2024-06-20
feat: add gzip of logs

Disservin:
When compiled with `ZLIB=true` and the `compress` option for the log is enabled, logs will be gzipped and get the .gz extension and a date timestmap.


Disservin:
for a typical log file this reduces the size by a factor of 6
```
-rw-rw-r--  1 max max  15M Jun 20 22:50 log
-rw-rw-r--  1 max max 2,5M Jun 20 22:55 log2024-06-20T.20.54.28.844102034.gz
```

gahtan-syarif:
how is the behavior when the log file already exists in the path? will it append to the original log file or overwrite it?

Disservin:
> how is the behavior when the log file already exists in the path? will it append to the original log file or overwrite it?

Already exists with the exact same timestamp ?

gahtan-syarif:
ahh right

vondele:
arguably that could be a feature for the pgn output as well

Disservin:
Yes I will add that later as well

PR #480, opened 2024-06-21
feat: add logic to get set options from engine

Disservin:


Disservin:
@gahtan-syarif lmk if this gives you everything

gahtan-syarif:
yeah this is enough for now i think

PR #489, opened 2024-06-23
fix: fastchess output for startpos

gahtan-syarif:


Disservin:
mh just leave it empty?

gahtan-syarif:
> mh just leave it empty?

looks kinda ugly ngl, extra comma and all:

Screenshot 2024-06-23 183045

i think it makes more sense to just put in startpos, so people know what opening is used

Disservin:
well you can also rewrite it so that `, ` is not shown..

gahtan-syarif:
> well you can also rewrite it so that `, ` is not shown..

well sure but that would complicate the code more than it needs to, i think just adding startpos is simpler and more informative

Disservin:
well the field name is "book name", if no book is given, the name should be empty and not "startpos" as that is not the name of the book

gahtan-syarif:
updated

Disservin:
what? you still have "startpos"

gahtan-syarif:
updated ver 2

Issue #477, opened 2024-06-21
report the fen and command when repoorting illegal moves


gahtan-syarif:
template example:
```
Warning; Illegal pv move b3f5 pv: info depth 16 seldepth 28 time 387 nodes 443259 nps 1142972 score cp -31 wdl 5 935 60 hashfull 5 pv e4f5 g6f5 f1e1 d6d5 c2c4 d5d4 f3h4 f5f4 h4f3 f4g3 f2g3 b3f5
From: position fen rnbqkbnr/pppppppp/8/8/4P3/8/PPPP1PPP/RNBQKBNR b KQkq e3 0 14 moves g6f5 f1e1 d6d5 c2c4 d5d4
Command: go wtime 232 btime 267
```

this is a requested feature in cutechess, and speeds up the debugging process as the user no longer has to check the log file.

PR #491, opened 2024-06-23
feat: show fen position when encountering an illegal pv move

Disservin:
closes https://github.com/Disservin/fast-chess/issues/477


Disservin:
I don't see much value in having the `Command` there

Issue #500, opened 2024-06-23
illegal pv warning prints wrong fen


gahtan-syarif:
illegal pv warning prints wrong fen
this is a match from startpos, it seems it prints the current fen of the board instead of the initial fen of the opening book
```
Warning; Illegal pv move c7c6 pv: info depth 11 seldepth 19 time 4 nodes 3238 nps 716641 score cp -8 wdl 13 962 25 hashfull 0 pv e8g8 d1g4 g8h8 g5e7 d8e7 c7c6 c7c5 g4g5 f7f6 e5f6 d7f6 d4c5 c8d7
From; position fen rnbqk2r/pppnbppp/4p3/3pP1B1/3P3P/2N5/PPP2PP1/R2QKBNR b KQkq - 0 6 moves e8g8 d1g4 g8h8 g5e7 d8e7
```
should be
```
Warning; Illegal pv move c7c6 pv: info depth 11 seldepth 19 time 4 nodes 3238 nps 716641 score cp -8 wdl 13 962 25 hashfull 0 pv e8g8 d1g4 g8h8 g5e7 d8e7 c7c6 c7c5 g4g5 f7f6 e5f6 d7f6 d4c5 c8d7
From; position startpos moves e8g8 d1g4 g8h8 g5e7 d8e7
```

gahtan-syarif:
it seems the moves are wrong too, it just takes the move of the pv instead of the played moves

Disservin:
Yeah I guess the position from the startpos is more useful

PR #501, opened 2024-06-23
fix: illegal pv warning

gahtan-syarif:
https://github.com/Disservin/fast-chess/issues/500

### DIFF
--- a/app/src/matchmaking/match/match.cpp
+++ b/app/src/matchmaking/match/match.cpp
@@ -58,7 +58,7 @@ void Match::addMoveData(const Player& player, int64_t measured_time_ms, bool leg
 
     move_data.score_string = ss.str();
 
-    verifyPvLines(player);
+    verifyPvLines(player, start_position_);
 
     data_.moves.push_back(move_data);
 }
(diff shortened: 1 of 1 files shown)


app/src/matchmaking/match/match.cpp
```diff
@@ -323,8 +323,8 @@ bool Match::isUciMove(const std::string& move) noexcept {
 }
 
 void Match::verifyPvLines(const Player& us) {
-    const static auto verifyPv = [](Board board, const std::vector<std::string>& tokens, std::string_view info) {
-        const auto fen = board.getFen();
+    const static auto verifyPv = [this](Board board, const std::vector<std::string>& tokens, std::string_view info) {
```
Disservin, line 326: you cant capture this because it is static

Disservin: and even then this won't work because the uci_moves are not correct...

gahtan-syarif: yeah, im trying to find a way to get the fen and moves from the command builder

Disservin:
https://github.com/Disservin/fast-chess/pull/502

PR #504, opened 2024-06-25
refactor: prevent simultaneous usage of tc and st

gahtan-syarif:


Disservin:
```
src/matchmaking/tournament/tournament_manager.cpp:114:99: warning: comparison of constant ‘1’ with boolean expression is always false [-Wbool-compare]
  114 |         if ((engine_configs_[i].limit.tc.time != 0 + engine_configs_[i].limit.tc.fixed_time != 0) > 1) {
```

yk that is why unit tests are helpful or even local compiles :p

PR #510, opened 2024-06-27
chore: get rid of redundant code

gahtan-syarif:
already exists here: https://github.com/Disservin/fast-chess/blob/e6288fc0613c12017d5c22fa4101d1335f72c21a/app/src/time/timecontrol.hpp#L56


Disservin:
i'd just remove the other one because it comes later?

gahtan-syarif:
sure

PR #515, opened 2024-07-03
feat: add WL/DD Ratio to fastchess output

gahtan-syarif:
i think this is a useful stat for measuring if a book leans more to being too unbalanced or too balanced, if WL/DD>1 then we can infer that the book leans more to being unbalanced while WL/DD<1 means a book leans more to being balanced.
looks like this:
```
Results of random_move_1 vs random_move_2 (2+0.02, NULL, NULL, openings.pgn):
Elo: -34.86 +/- 62.20, nElo: -122.84 +/- 215.34
LOS: 13.18 %, DrawRatio: 80.00 %, PairsRatio: 0.00
Games: 10, Wins: 1, Losses: 2, Draws: 7, Points: 4.5 (45.00 %)
Ptnml(0-2): [0, 1, 4, 0, 0], WL/DD Ratio: 0.33
```

### DIFF
--- a/app/src/matchmaking/output/output_fastchess.hpp
+++ b/app/src/matchmaking/output/output_fastchess.hpp
@@ -85,7 +85,7 @@ class Fastchess : public IOutput {
             auto line5 = fmt::format("Ptnml(0-2): [{}, {}, {}, {}, {}], WL/DD Ratio: {:.2f}", stats.penta_LL, stats.penta_LD,
                                      stats.penta_WL + stats.penta_DD, stats.penta_WD, stats.penta_WW, WLDDRatio);
 
-            return fmt::format("{}\n{}\n{}\n{}\n", line1, line2, line3, line4, line5);
+            return fmt::format("{}\n{}\n{}\n{}\n{}\n", line1, line2, line3, line4, line5);
         }
 
         return fmt::format("{}\n{}\n{}\n{}\n", line1, line2, line3, line4);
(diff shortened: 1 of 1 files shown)


app/src/matchmaking/output/output_fastchess.hpp
```diff
@@ -79,9 +80,13 @@ class Fastchess : public IOutput {
...
-        auto lines = fmt::format("{}\n{}\n{}\n{}\n{}\n", line1, line2, line3, line4, line5);
+        auto line5 = fmt::format("Ptnml(0-2): [{}, {}, {}, {}, {}], WL/DD Ratio: {:.2f}", stats.penta_LL, stats.penta_LD,
+                                 stats.penta_WL + stats.penta_DD, stats.penta_WD, stats.penta_WW, WLDDRatio);
+        std::string lines;
+        if (report_penta_)
+            lines = fmt::format("{}\n{}\n{}\n{}\n{}\n", line1, line2, line3, line4, line5);
+        else
+            lines = fmt::format("{}\n{}\n{}\n{}\n", line1, line2, line3, line4);
 
         return lines;
```
Disservin, line 91: you could just return early
```cpp
if (report_penta_) {
        auto line5 = fmt::format("Ptnml(0-2): [{}, {}, {}, {}, {}], WL/DD Ratio: {:.2f}", stats.penta_LL, stats.penta_LD,
                                 stats.penta_WL + stats.penta_DD, stats.penta_WD, stats.penta_WW, WLDDRatio);

        return fmt::format("{}\n{}\n{}\n{}\n", line1, line2, line3, line4, line5);
}

return fmt::format("{}\n{}\n{}\n{}\n", line1, line2, line3, line4);
```

Disservin (changes requested):

app/src/matchmaking/output/output_fastchess.hpp
gahtan-syarif: fixed 👍

PR #516, opened 2024-07-05
fixed-time-control test for e2e

gahtan-syarif:


Disservin:
can you maybe also check the time in the pgn, the current state is perhaps a bit useless

gahtan-syarif:
u mean the pgnout output from the ci? how to do that?

Disservin:
In the script check that the times in the pgn are what you expected ?

gahtan-syarif:
hmm the random mover doesnt seem to have a time control logic, so this pr is useless

Issue #517, opened 2024-07-05
todo: fix weird output


gahtan-syarif:
![Screenshot 2024-07-05 135612](https://github.com/Disservin/fast-chess/assets/155860115/2f4f8991-a34f-4a66-81d4-0122a0d5cf79)

Disservin:
https://github.com/Disservin/fast-chess/pull/518

PR #518, opened 2024-07-05
fix occasional weird output

gahtan-syarif:
fix issue https://github.com/Disservin/fast-chess/issues/517

### DIFF
--- a/app/src/matchmaking/output/output_fastchess.hpp
+++ b/app/src/matchmaking/output/output_fastchess.hpp
@@ -20,11 +20,11 @@ class Fastchess : public IOutput {
 
     void printInterval(const SPRT& sprt, const Stats& stats, const std::string& first, const std::string& second,
                        const engines& engines, const std::string& book) override {
-        std::cout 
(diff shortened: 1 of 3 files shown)


app/src/matchmaking/output/output_fastchess.hpp
```diff
@@ -20,17 +20,18 @@ class Fastchess : public IOutput {
...
                        const engines& engines, const std::string& book) override {
-        std::cout << "--------------------------------------------------\n" << std::flush;
-        printElo(stats, first, second, engines, book);
-        printSprt(sprt, stats);
-        std::cout << "--------------------------------------------------\n" << std::flush;
+        std::cout <<
+        "--------------------------------------------------\n" <<
+        printElo(stats, first, second, engines, book) <<
+        printSprt(sprt, stats) <<
+        "--------------------------------------------------\n" << std::flush;
```
Disservin, line 27: this still races.. you can't use multiple `<<` make it one string and then output @gahtan-syarif

PR #519, opened 2024-07-05
fix: make sprt result format consistent with cutechess output

gahtan-syarif:
looks like this for cutechess output:
```
SPRT: llr 2.98 (101.2%), lbound -2.94, ubound 2.94 - H1 was accepted
```

### DIFF
--- a/app/src/matchmaking/output/output_cutechess.hpp
+++ b/app/src/matchmaking/output/output_cutechess.hpp
@@ -35,8 +35,22 @@ class Cutechess : public IOutput {
 
     std::string printSprt(const SPRT& sprt, const Stats& stats) override {
         if (sprt.isEnabled()) {
-            auto fmt = fmt::format("LLR: {:.2f} {} {}\n", sprt.getLLR(stats.wins, stats.draws, stats.losses),
-                                   sprt.getBounds(), sprt.getElo());
+            double lowerBound = sprt.getLowerBound();
+            double upperBound = sprt.getUpperBound();
+            double llr = sprt.getLLR(stats.wins, stats.draws, stats.losses);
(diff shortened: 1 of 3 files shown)


app/src/matchmaking/output/output_cutechess.hpp
```diff
@@ -35,8 +35,22 @@ class Cutechess : public IOutput {
...
+            double percentage = llr < 0 ? llr / lowerBound * 100 : llr / upperBound * 100;
+           
+            std::string result;
+            if (llr >= upperBound) {
+                result = " - H1 was accepted";
+            } else if (llr < lowerBound) {
+                result = " - H0 was accepted";
+            } else {
+                result = "";
+            }
```
Disservin, line 50: this is useless?

Disservin (changes requested):

app/src/matchmaking/output/output_cutechess.hpp
```diff
@@ -35,8 +35,22 @@ class Cutechess : public IOutput {
...
+           
+            std::string fmt = fmt::format("SPRT: llr {.2f} ({.1f}%), lbound {}, ubound {}{}\n",
```
Disservin, line 52: also just return fmt::format directly here

app/src/matchmaking/output/output_cutechess.hpp
gahtan-syarif: removed 👍

app/src/matchmaking/output/output_cutechess.hpp
gahtan-syarif: fixed 👍

PR #521, opened 2024-07-05
add logger warning for disconnect

gahtan-syarif:


app/src/matchmaking/match/match.cpp
```diff
@@ -266,8 +266,11 @@ void Match::setEngineCrashStatus(Player& loser, Player& winner) {
 
     crash_or_disconnect_ = true;
 
+    const auto name = loser.engine.getConfig().name;
     data_.termination = MatchTermination::DISCONNECT;
     data_.reason      = loser.engine.getConfig().name + Match::DISCONNECT_MSG;
```
Disservin, line 271: use the name var here as well?

Disservin (changes requested):

app/src/matchmaking/match/match.cpp
gahtan-syarif: ahh right

gahtan-syarif: updated 👍

Issue #528, opened 2024-07-09
parsing options


vondele:
fishtest sends this option to cutechess-cli / fast-chess:
```
./fast-chess    -recover -repeat -games 466 -tournament gauntlet\
                -pgnout results-a181cd67-43a0-4d6e-b6b4-98f9b5fa7e7e.pgn\
                -site 'https://tests.stockfishchess.org/tests/view/66872b0ffe7b81f5e163473c' -event 'Batch 295: master vs master'\
                -srand $RANDOM  -resign movecount=3 score=600 -draw movenumber=34 movecount=8 score=20\
                -variant standard -concurrency 4 -openings file=UHO_Lichess_4852_v1.epd format=epd order=random plies=16 start=52569\
                -engine name=New-ee6fc7e38b4aeef44862159215a56d97122f59a0 tc=0.539+0.005 cmd=./stockfish_ee6fc7e38b4aeef44862159215a56d97122f59a0 dir=.\
                        option.Hash=16 option.EvalFile=nn-ddcfb9224cdb.nnue option.EvalFileSmall=nn-37f18f62d772.nnue\
                -engine name=Base-ee6fc7e38b4aeef44862159215a56d97122f59a0 tc=0.539+0.005 cmd=./stockfish_ee6fc7e38b4aeef44862159215a56d97122f59a0 dir=.\
                        option.Hash=16 option.EvalFile=nn-ddcfb9224cdb.nnue option.EvalFileSmall=nn-37f18f62d772.nnue\
                -each proto=uci option.Threads=1 
```
works with cutechess, but fast-chess has:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  
Error while reading option "-engine" with value "option.EvalFileSmall=nn-37f18f62d772.nnue"
Reason: Error; Engine not found: 

```

gahtan-syarif:
maybe because of the dash in the filename?

gahtan-syarif:
or on second thought its probably the dot
```cpp
        // Strip option.Name of the option. Part
        const std::size_t pos         = key.find('.');
        const std::string strippedKey = key.substr(pos + 1);
```

vondele:
`std::find_first_of`  ?

gahtan-syarif:
wait, weirdly enough it works for me:
```
[...]

PR #530, opened 2024-07-09
fix: missing parenthesis on validateEnginePath

gahtan-syarif:
missing parenthesis for the engine path validation https://github.com/Disservin/fast-chess/blob/2a4ffaf57857df102dad4a1efb51106190fbd0c7/app/src/engine/uci_engine.cpp#L183

issue https://github.com/Disservin/fast-chess/issues/528


gahtan-syarif:
@vondele does this fix it?

vondele:
yes

vondele:
hmm, not sure, I have 'Warning; Engine New-ee6fc7e38b4aeef44862159215a56d97122f59a0 disconnects' .. might be related to this.

PR #543, opened 2024-07-12
disable realtime_logging by default for windows

gahtan-syarif:


Disservin:
not a fan i dislike ahving different defaults depending on the os

PR #558, opened 2024-07-13
refactor: get rid of validateEnginePath

gahtan-syarif:


gahtan-syarif:
> id still like to have before hand cause and not later

the current way has problem where it will throw error when `dir=` is placed after `-each` instead of `-engine` , by placing it afterhand we ensure that the dir and cmd variables are already populated before validation

Disservin:
> > id still like to have before hand cause and not later
> 
> the current way has problem where it will throw error when `dir=` is placed after `-each` instead of `-engine` , by placing it afterhand we ensure that the dir and cmd variables are already populated before validation

then move it into the sanitize part

Disservin:
here https://github.com/Disservin/fast-chess/blob/master/app/src/config/sanitize.cpp

gahtan-syarif:
done 👍 

> here https://github.com/Disservin/fast-chess/blob/master/app/src/config/sanitize.cpp

PR #563, opened 2024-07-14
refactor: make output exactly the same as cutechess

gahtan-syarif:
improve cutechess compatibility

### DIFF
--- a/app/src/book/opening_book.cpp
+++ b/app/src/book/opening_book.cpp
@@ -6,6 +6,7 @@
 
 #include <config/config.hpp>
 #include <util/safe_getline.hpp>
+#include <util/logger/logger.hpp>
 
 namespace fast_chess::book {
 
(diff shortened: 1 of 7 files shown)


vondele:
There are two more differences remaining:

just naming:
```
< Finished game 1 (sf_1 vs sf_2): 1/2-1/2 {Draw by insufficient mating material}
---
> Finished game 1 (sf_1 vs sf_2): 1/2-1/2 {Draw by insufficient material}
```

The other one is: 
```
< Finished game 2 (sf_2 vs sf_1): 1/2-1/2 {Draw by 3-fold repetition}
---
> Finished game 2 (sf_2 vs sf_1): 1/2-1/2 {Draw by adjudication}
```

Here it looks like in cutechess the actual game status is checked first, while the adjudication comes afterwards. (The game does end in a real 3-fold).

PR #569, opened 2024-07-14
refactor: print "Finished game ..." alongside the scoreline to prevent chaotic output

gahtan-syarif:
prevent cases where finished game is printed twice before the scoreline, as this may affect fishtest parsing.

prevents this:
```
Finished game 64 (Engine2 vs Engine1): 1-0 {Black loses on time}
Finished game 63 (Engine1 vs Engine2): 1-0 {Black loses on time}
Score of Engine1 vs Engine2: 38 - 25 - 0  [0.603] 63
Started game 67 of 100 (Engine1 vs Engine2)
Score of Engine1 vs Engine2: 39 - 25 - 0  [0.609] 64
```

### DIFF
--- a/app/src/matchmaking/tournament/roundrobin/roundrobin.cpp
+++ b/app/src/matchmaking/tournament/roundrobin/roundrobin.cpp
@@ -65,7 +65,7 @@ void RoundRobin::create() {
         // callback functions, do not capture by reference
         const auto finish = [this, configs, first, second, game_id, round_id](
                                 const Stats& stats, const std::string& reason, const engines& engines) {
-            output_->endGame(configs, stats, reason, game_id);
+            std::string finish_message = output_->endGame(configs, stats, reason, game_id);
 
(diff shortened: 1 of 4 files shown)


vondele:
I can confirm that with that patch the fishtest worker runs without asserts triggered (while with master it still does).

app/src/matchmaking/tournament/roundrobin/roundrobin.cpp
```diff
@@ -75,6 +73,8 @@ void RoundRobin::create() {
             // Score of Engine1 vs Engine2: 95 - 92 - 0  [0.508] 187
             // Score of Engine1 vs Engine2: 94 - 92 - 0  [0.505] 186
             std::lock_guard<std::mutex> lock(output_mutex_);
+            
+            std::string finish_message = output_->endGame(configs, stats, reason, game_id);
```
Disservin, line 77: you shouldnt need returning strings here if you move it after the mutex then there is no racy behavior

Issue #572, opened 2024-07-15
Throw error when opening book seed is not specified


gahtan-syarif:
i notice engine devs have a bad habit of not specifying the book seed or randomseed on tests. theres currently a warning for it but they tend to ignore this. making a pr soon

robertnurnberg:
Wouldn't this go against the spirit of having order default to random. That would rather suggest to use randomseed as default, too. It's a design decision. Maybe now is the right time to think carefully of what is best. Options are to leave master as is, follow exactly what cute-chess is doing, or mandating order and seed to be specified.

gahtan-syarif:
my thinking is to make the seed specification mandatory only when the order is random

robertnurnberg:
And random is currently the default? Hence my point.

Disservin:
What is wrong with not specifying one?

gahtan-syarif:
> What is wrong with not specifying one?

concerns of overfitting to the first N openings of the book, as with the default seed the opening sequences doesnt change after each test. also having opening book seed randomized by default makes the behavior consistent with cutechess, which is necessary for the cutechess build

Issue #546, opened 2024-07-12
opening order default


vondele:
Openings will be picked in the order specified by ORDER,
			which can be either 'random' or 'sequential' (default).

vs 

               ORDER  - the order of openings to be played, either random or sequential.
                         Default is random.

robertnurnberg:
Noting here that if default behaviour is changed, then parts of https://github.com/Disservin/fast-chess/pull/448 should be reverted again.

gahtan-syarif:
i think random for default makes more sense as its the one testers typically use, and fishtest also already specifies `order=random`

vondele:
I'm find with any default as long as the choice is conscious.

PR #573, opened 2024-07-15
refactor: set default order to sequential and remove -randomseed

gahtan-syarif:
consistent with cutechess
fix https://github.com/Disservin/fast-chess/issues/572
and
fix https://github.com/Disservin/fast-chess/issues/546


app/src/cli/cli.cpp
```diff
@@ -487,13 +488,8 @@ void parseRecover(const std::vector<std::string> &, ArgumentData &argument_data)
...
 }
 
-void parseRandomSeed(const std::vector<std::string> &, ArgumentData &argument_data) {
-    argument_data.tournament_config.randomseed = true;
-
-    std::random_device rd;
-    std::mt19937_64 gen((static_cast<uint64_t>(rd()) << 32) | rd());
-    std::uniform_int_distribution<uint64_t> dist(0, std::numeric_limits<uint64_t>::max());
-    argument_data.tournament_config.seed = dist(gen);
+void parseRandomSeed(const std::vector<std::string> &, ArgumentData &) {
```
robertnurnberg, line 491: I think this can now be removed, together with the line that calls this function?

robertnurnberg:
As part of this PR I would suggest to adjust the change from https://github.com/Disservin/fast-chess/pull/551, and make fast-chess always behave the same as far as the inputs are concerned, whether in cute mode or not. For example, let the order be random by default, but throw an error when `-order` is not specified in cute mode. This can be achieved by setting a boolean `read_order_success` in cute mode in `cli.cpp`.

app/src/cli/cli.cpp
gahtan-syarif: I'd rather keep it there so if someone specifies '-randomseed' it will just be silently ignored instead of throwing an error

gahtan-syarif:
> As part of this PR I would suggest to adjust the change from #551, and make fast-chess always behave the same as far as the inputs are concerned, whether in cute mode or not. For example, let the order be random by default, but throw an error when `-order` is not specified in cute mode. This can be achieved by setting a boolean `read_order_success` in cute mode in `cli.cpp`.

I see, for this id like to make it a seperate pr

Disservin:
i don't like this.. id prefer a deterministic tournament config/setup unless the user actively specifies it differently

gahtan-syarif:
> i don't like this.. id prefer a deterministic tournament config/setup unless the user actively specifies it differently

perhaps order=sequential as default? because i agree with robert that if someone specifies order=random they typically want it to be non-deterministic and already have the seed randomized

gahtan-syarif:
made it so that the default order is sequential so default behavior will still be deterministic

robertnurnberg:
> i don't like this.. id prefer a deterministic tournament config/setup unless the user actively specifies it differently

Would you be happy to have sequential order as default then? Would be consistent with cute chess and do what you want.

At present cute chess and fast chess behave fundamentally differently for the same input options. That is not ideal.

gahtan-syarif:
i really think you should consider adding this, rn the only way to get a randomized seed in fastchess is through the use of `-randomseed` , but this isnt available in native cutechess, so tools who will use fastchess' cutechess build will still have to revise their command builder. this defeats the purpose of the cutechess build

app/src/cli/cli.cpp
robertnurnberg: I think this PR has moved to mirror cute-chess, which I think is good. So then we should also remove this. It's a simple change for any user getting caught out.

gahtan-syarif: removed for now, but might add a logger warning in the future that the option has been removed if someone asks why the option throws an error

app/src/book/opening_book.cpp
```diff
@@ -58,7 +58,11 @@ void OpeningBook::setup(const std::string& file, FormatType type) {
         shuffle();
     }
 
-    rotate(offset_);
+    if (offset_ > 0) {
+        Logger::trace("Applying opening book offset...");
+        rotate(offset_);
```
robertnurnberg, line 63: Could output value of offset here?

gahtan-syarif: done

app/src/config/sanitize.cpp
```diff
@@ -5,13 +5,23 @@
...
 
 namespace fast_chess::config {
 
 void sanitize(config::Tournament& config) {
+    if (config.randomize_seed && config.opening.order == OrderType::RANDOM) {
+        std::random_device rd;
+        std::mt19937_64 gen((static_cast<uint64_t>(rd()) << 32) | rd());
+        std::uniform_int_distribution<uint64_t> dist(0, std::numeric_limits<uint64_t>::max());
+        config.seed = dist(gen);
+        // make sure that seed isnt randomized when saved and loaded from config
```
robertnurnberg, line 21: is not

gahtan-syarif: done

gahtan-syarif:
@Disservin so whats the status with this pr? do i just close it or is there something i need to fix?

PR #574, opened 2024-07-15
refactor: set default log level to trace

gahtan-syarif:
i think trace provides a lot of useful information for debugging, and its what most of the log consists of, so i feel its better to set it as the default level


gahtan-syarif:
could do with an explanation why...

Disservin:
It is generally bad practice, if something should be visible earlier just change the log level

Issue #577, opened 2024-07-15
different game lengths


vondele:
It looks like the PlyCount can be different between cutechess and fastchess as observed https://github.com/Disservin/fast-chess/pull/570#issuecomment-2228943498

```
0a1
> 14
2c3,4
< 15
---
> 18
> 18
9a12
> 20
17,21c20,22
< 23
< 23
< 23
< 23
< 23
---
> 22
> 22
> 22
24,26c25,27
< 25
< 25
< 25
---
```

Obtained with:
```bash
rm -f *-out.pgn

for binary in ./cutechess-cli ./fast-chess
do

echo " ============= $binary ============== "

(

$binary         -recover -repeat -games 2 -rounds 100 -tournament gauntlet\
                -pgnout $binary-out.pgn\
                -srand $RANDOM  -resign movecount=3 score=600 -draw movenumber=34 movecount=8 score=20\
                -variant standard -concurrency 25 -openings file=book.epd format=epd order=sequential \
                -engine name=sf_1 tc=100+0.1 depth=6 cmd=./sf_1 dir=.\
                -engine name=sf_2 tc=100+0.1 depth=8 cmd=./sf_2 dir=.\
                -each proto=uci option.Threads=1

) | grep "Finished game" | sort -n -k 3 > $binary-out.finished

done

echo "============== Finished game output ================"
diff ./cutechess-cli-out.finished ./fast-chess-out.finished
echo "============== pgn result =========================="
grep PlyCount cutechess-cli-out.pgn | grep -o "[0-9]*" | sort -n > cutechess-cli-out.plycount
grep PlyCount fast-chess-out.pgn | grep -o "[0-9]*" | sort -n > fast-chess-out.plycount
diff ./cutechess-cli-out.plycount ./fast-chess-out.plycount 
echo "===================================================="
```

vondele:
Is there any interest in having the above script as an action that runs on the fast-chess repo? 

If so, I can try to make something, let me know.

Disservin:
yes please :D

gahtan-syarif:
i found the issue, the latest commit seems to have messed up the resigntracker, ill work on a fix

PR #579, opened 2024-07-15
fix: different game length between cutechess and fastchess

gahtan-syarif:
fix https://github.com/Disservin/fast-chess/issues/577


gahtan-syarif:
this should fix it

gahtan-syarif:
@vondele this should fix it

Disservin:
> @vondele this should fix it

no

gahtan-syarif:
> > this should fix it
> 
> no

it does, it fixed the game from adjudicating one ply late

Disservin:
ah wait had the wrong book whops

PR #590, opened 2024-07-15
refactor: remove unecessary conditional

gahtan-syarif:
no longer needed now that checkmates are prioritized


Disservin:
can you drop a comment stating that ? just want it future proof in case something gets reordered or is unclear

gahtan-syarif:
added comment

PR #599, opened 2024-07-16
fix: don't wait for a read call to return before terminating the threadpool

Disservin:


app/src/util/threadpool.hpp
```diff
@@ -66,6 +66,11 @@ class ThreadPool {
 
         for (auto &worker : workers_) {
             if (worker.joinable()) {
+#ifdef _WIN65
```
gahtan-syarif, line 69: win65?

Disservin: kekw

Issue #562, opened 2024-07-13
Hardcode source and header files in the makefile


gahtan-syarif:
needs for fix for fishtest integration https://github.com/official-stockfish/fishtest/pull/2119

PR #608, opened 2024-07-18
refactor: remove find command and hardcode source and header files in the makefile

gahtan-syarif:
successfully compiles
fix https://github.com/Disservin/fast-chess/issues/562

### DIFF
--- a/app/Makefile
+++ b/app/Makefile
@@ -264,27 +264,6 @@ build_directories:
 	@$(MKDIR) $(BUILDDIR)/tests/e2e
 	@$(MKDIR) $(BUILDDIR)/tests/"mock/engine"
 	@$(MKDIR) $(BUILDDIR)/third_party
-	@$(MKDIR) $(BUILDDIR)/third_party/fmt
-	@$(MKDIR) $(BUILDDIR)/third_party/fmt/".github"
-	@$(MKDIR) $(BUILDDIR)/third_party/fmt/".github"/workflows
-	@$(MKDIR) $(BUILDDIR)/third_party/fmt/doc
-	@$(MKDIR) $(BUILDDIR)/third_party/fmt/"include/fmt"
-	@$(MKDIR) $(BUILDDIR)/third_party/fmt/src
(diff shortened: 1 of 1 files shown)


app/Makefile
```diff
@@ -134,9 +233,59 @@ $(BUILDDIR)/%.o: %.cpp | build_directories
...
+	@$(MKDIR) $(BUILDDIR)/third_party/fmt/test
+	@$(MKDIR) $(BUILDDIR)/third_party/fmt/test/add-subdirectory-test
+	@$(MKDIR) $(BUILDDIR)/third_party/fmt/test/compile-error-test
+	@$(MKDIR) $(BUILDDIR)/third_party/fmt/test/cuda-test
+	@$(MKDIR) $(BUILDDIR)/third_party/fmt/test/find-package-test
+	@$(MKDIR) $(BUILDDIR)/third_party/fmt/test/fuzzing
+	@$(MKDIR) $(BUILDDIR)/third_party/fmt/test/gtest
+	@$(MKDIR) $(BUILDDIR)/third_party/fmt/test/gtest/gmock
+	@$(MKDIR) $(BUILDDIR)/third_party/fmt/test/gtest/gtest
+	@$(MKDIR) $(BUILDDIR)/third_party/fmt/test/static-export-test
```
Disservin, line 287: you dont need this, ony "gzip" dir is required

app/Makefile
```diff
@@ -134,9 +233,59 @@ $(BUILDDIR)/%.o: %.cpp | build_directories
 
 build_directories:
```
Disservin, line 234: mh ill have to look if there is a better way to solve this

Disservin (changes requested):

app/Makefile
gahtan-syarif: removed 👍

app/Makefile
gahtan-syarif: in the meantime i think its good to merge this first, so we can go ahead with the fishtest integration

PR #615, opened 2024-07-20
refactor: more logs

Disservin:


app/src/matchmaking/tournament/base/tournament.cpp
```diff
@@ -81,24 +81,27 @@ void BaseTournament::playGame(const GamePair<EngineConfiguration, EngineConfigur
...
 
-    Logger::trace("Playing game {} between {} and {}", game_id, white_name, black_name);
+    Logger::trace<true>("Game {} between {} and {} starting", game_id, white_name, black_name);
 
     start();
 
     auto match = Match(opening);
     match.start(white_engine.get(), black_engine.get(), core.get().cpus);
 
+    Logger::trace<true>("Game {} between {} and {} finished", game_id, white_name, black_name);
```
robertnurnberg, line 91: Would it make sense to include here how it ended? And below which engine crashed or disconnected? Although the crash/disconnect is maybe itself already in the log before, not sure.

PR #616, opened 2024-07-20
fix: seed not getting randomized

gahtan-syarif:
@Disservin this does not work https://github.com/Disservin/fast-chess/pull/573/commits/22f57c22c74574b08ea3424bb55a6077ed1818de the randomized seed just gets overwritten by the default seed in types/tournament.hpp


Disservin:
i dont want this variable in the struct, it isnt needed

Disservin:
ill fix it

PR #621, opened 2024-07-21
make fast-chess spelling consistent

robertnurnberg:
I checked any hits on `grep -iRl --include=\*.{txt,md,c,cpp,py,h,hpp} --include=man "fastchess" .` and `find . -type f -iname '*fastchess*'`, leaving internal variable names unchanged.

For the (non)-capitalization of `Fast-chess` in `man` I checked output of `man man`.

This PR brings fast-chess itself in line with the consistent spelling in https://github.com/official-stockfish/fishtest/pull/2119.


Disservin:
mh internally i think id still like to refer to it as fastchess, easier to type than ahving to add a hyphen in between

Disservin:
but tbh i havent thought much about the name actually

vondele:
not too late to change the name tbh. The hypen is indeed a bit annoying at times, it requires to use two different spellings in code.

Disservin:
tbh i only later realized that there already exists an engine with this name https://github.com/thomasahle/fastchess, avoiding some confusion/ a more unique name would be nice :p

robertnurnberg:
Are we open to completely new names as well then? Does a quickchess exist?

gahtan-syarif:
I think for a complete rename we need to ask the opinions of PGG106 and Szil as well as primary authors

Disservin:
i dont think ive ever asked them about the name, was rather something i forced upon them

gahtan-syarif:
> i dont think ive ever asked them about the name, was rather something i forced upon them

Then they might have other ideas for the name in that case, could use their opinion as well :)

gahtan-syarif:
PGG106 suggests frog-chess, he might be joking im unsure

Disservin:
should name it mercury 

> Mercury is named after the messenger for their gods

afterall engines are godlike and we carry their message

gahtan-syarif:
sounds good, the roman god mercury is known as being very fast

vondele:
And not unrelated to chess https://en.wikipedia.org/wiki/Ca%C3%AFssa#Vida's_poem

robertnurnberg:
Not to forget that another name for mercury (the element) is *quick*silver. :)

robertnurnberg:
Anyway, if we go down that route and choose a new name, would it make sense to immediately transfer to a new repo under official-stockfish?

In the meantime, I will close this PR.

Disservin:
I don't see myself moving it to the stockfish org tbh, as of now I'm maintaining it independently

gahtan-syarif:
quicksilver sounds pretty cool

Issue #623, opened 2024-07-21
fast-chess fails to extract bestmove from python engines


robertnurnberg:
Example engines where this happens are https://github.com/Disservin/python-chess-engine and `cdb2uci.py` from https://github.com/robertnurnberg/cdblib.

With the latter wrapped in `dumb2uci.sh` as follows:
```
#! /bin/bash

python $HOME/git/cdblib/cdb2uci.py -c 1
```

I get these results with cutechess-cli and fast-chess:
```
> ./cutechess-cli -engine cmd=../../stockfish/src/stockfish name=sf tc=1+0.1 -engine cmd=../../cdblib/dumb2uci.sh name=cdb2uci tc=10+5 -each proto=uci -concurrency 1 -openings file=grob.epd -rounds 2 -games 1
Started game 1 of 2 (sf vs cdb2uci)
Finished game 1 (sf vs cdb2uci): 1-0 {White mates}
Score of sf vs cdb2uci: 1 - 0 - 0  [1.000] 1
Started game 2 of 2 (cdb2uci vs sf)
Finished game 2 (cdb2uci vs sf): 0-1 {Black mates}
Score of sf vs cdb2uci: 2 - 0 - 0  [1.000] 2
...      sf playing White: 1 - 0 - 0  [1.000] 1
...      sf playing Black: 1 - 0 - 0  [1.000] 1
...      White vs Black: 1 - 1 - 0  [0.500] 2
Elo difference: inf +/- nan, LOS: 92.1 %, DrawRatio: 0.0 %
SPRT: llr 0 (0.0%), lbound -inf, ubound inf

Player: sf
   "Win: Black mates": 1
   "Win: White mates": 1
Player: cdb2uci
   "Loss: Black mates": 1
   "Loss: White mates": 1
Finished match
```

```
> rm -f cdblog.txt && ./fast-chess -engine cmd=../stockfish/src/stockfish name=sf tc=1+0.1 -engine cmd=../cdblib/cdb2uci.sh name=cdb2uci tc=10+5 -concurrency 1 -openings file=grob.epd -log file=cdblog.txt level=trace -rounds 2 -games 1 -pgnout file=grob_sf_cdb2uci.pgn
Started game 1 of 2 (sf vs cdb2uci)
Warning; No bestmove found in the last line from cdb2uci
Warning; No bestmove found in the last line from cdb2uci
Warning; Engine cdb2uci loses on time
[...]

PR #629, opened 2024-07-21
feat: check if engine was properly started

Disservin:
fails now on the example from https://github.com/Disservin/fast-chess/issues/623

### DIFF
--- a/app/src/engine/process/process_posix.hpp
+++ b/app/src/engine/process/process_posix.hpp
@@ -44,49 +44,6 @@ namespace engine::process {
 
 inline char **environ;
 
-class Pipes {
-    int pipe_[2];
-
-    int open_ = -1;
-
-   public:
-    Pipes() {
-        if (pipe(pipe_) == -1) {
-            throw std::runtime_error("Failed to create pipe.");
-        }
-    }
-
-    void dup2(int fd, int fd2) {
-        if (::dup2(pipe_[fd], fd2) == -1) {
-            throw std::runtime_error("Failed to duplicate pipe.");
-        }
-
-        close(pipe_[fd]);
-
-        // 1 -> 0, 0 -> 1
-        open_ = 1 - fd;
-    }
-
-    void setNonBlocking() const noexcept { fcntl(get(), F_SETFL, fcntl(get(), F_GETFL) | O_NONBLOCK); }
-
-    void setOpen(int fd) noexcept { open_ = fd; }
-
-    [[nodiscard]] int get() const noexcept {
-        assert(open_ != -1);
-        return pipe_[open_];
-    }
-
-    [[nodiscard]] int get(int fd) const noexcept {
-        assert(open_ != -1);
-        return pipe_[fd];
-    }
-
-    ~Pipes() {
-        close(pipe_[0]);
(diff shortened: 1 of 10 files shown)


robertnurnberg:
Confirmed behaviour as desired:
```
> ./fast-chess -engine cmd=../cdblib/cdb2uci.sh name=cdb1 -engine cmd=../cdblib/cdb2uci.sh name=cdb2 -each tc=10+5 -concurrency 1 -openings file=grob.epd -rounds 2 -games 2
Started game 1 of 4 (cdb1 vs cdb2)
Warning: Cannot start engine cdb1:
Cannot execute command: ../cdblib/cdb2uci.sh
Finished game 1 (cdb1 vs cdb2): 1/2-1/2 {}
Finished match
> cat ../cdblib/cdb2uci.sh
python $HOME/git/cdblib/cdb2uci.py -c 1 -ee
```

app/src/engine/process/process_posix.hpp
```diff
@@ -121,18 +143,33 @@ class Process : public IProcess {
...
+            const std::string status_file = "/status";
+            const auto path               = base_path + std::to_string(fork_pid) + status_file;
+
+            while (!fileReady(path)) {
+                std::this_thread::sleep_for(std::chrono::milliseconds(10));
+            }
+
+            // wait until the process is spawned and the name has been changed away from
+            // Name:   fast-chess to the actual name of the process
+            if (isZombie(path, "fast-chess")) {
```
vondele, line 163: This will mean users can't change the name of the binary (like fast-chess.master, fast-chess.patch, which is common). You might be able to get it from argv 0 ?

Disservin: good point, will change that later and use something like this to detect if the process is working, since the proc/pid/status approach might not work for other platforms

```cpp
#include <stdio.h>
#include <sys/sysctl.h>
#include <sys/types.h>
#include <sys/user.h>
#include <unistd.h>
#include <stdlib.h>

int main() {
    int mib[4];
    size_t len;
    struct kinfo_proc *procs;
    int proc_count;

    mib[0] = CTL_KERN;
    mib[1] = KERN_PROC;
    mib[2] = KERN_PROC_ALL;
    mib[3] = 0;

    // First call to sysctl to get the size of the data to be returned
    sysctl(mib, 4, NULL, &len, NULL, 0);
    procs = (struct kinfo_proc *)malloc(len);

    // Second call to sysctl to get the actual process data
    sysctl(mib, 4, procs, &len, NULL, 0);

    proc_count = len / sizeof(struct kinfo_proc);
    for (int i = 0; i < proc_count; i++) {
        struct kinfo_proc *proc = &procs[i];

#ifdef __FreeBSD__
        if (proc->ki_stat == SZOMB) {
            printf("Process %d is a zombie\n", proc->ki_pid);
        }
#else
        if (proc->kp_proc.p_stat == SZOMB) {
            printf("Process %d is a zombie\n", proc->kp_proc.p_pid);
        }
#endif
    }

    free(procs);
    return 0;
}
```

vondele: yeah, I think this is rather system dependent and thus fragile code.

PR #630, opened 2024-07-21
refactor: add "normal" termination tag for pgn

gahtan-syarif:
this is part of the pgn standard http://www.saremba.de/chessgml/standards/pgn/pgn-complete.htm#c9.8.1 goal is to make it easier to filter out normal matches with adjudications/disconnects/illegal moves etc...


Disservin:
> normal matches with adjudications/disconnects/illegal moves etc...

This doesnt make sense? They already have a different termination header

PR #633, opened 2024-07-22
fix: apply increment before the first move

gahtan-syarif:
https://github.com/cutechess/cutechess/blob/780065637f9936bc29cc592c6f0b2007ccbf66de/projects/lib/src/timecontrol.cpp#L221

its also part of fide rules ex. https://www.youtube.com/watch?v=d-7ZpPDrOOo

### DIFF
--- a/app/src/matchmaking/player.hpp
+++ b/app/src/matchmaking/player.hpp
@@ -14,9 +14,7 @@ class Player {
     // The timeout threshold for the read engine command.
     // This has nothing to do with the time control itself.
     [[nodiscard]] std::chrono::milliseconds getTimeoutThreshold() const {
-        if (engine.getConfig().limit.nodes != 0     //
-            || engine.getConfig().limit.plies != 0  //
-            || time_control_.isFixedTime()) {
+        if (engine.getConfig().limit.nodes != 0 || engine.getConfig().limit.plies != 0) {
             // no timeout
             return std::chrono::milliseconds(0);
         }
--- a/app/src/time/timecontrol.cpp
+++ b/app/src/time/timecontrol.cpp
@@ -16,10 +16,8 @@ namespace fast_chess {
 TimeControl::TimeControl(const Limits &limits) : limits_(limits) {
     if (limits_.fixed_time != 0) {
         time_left_ = limits_.fixed_time;
-    } else if (limits_.time != 0) {
-        time_left_ = limits_.time + limits_.increment;
     } else {
-        time_left_ = 0;
+        time_left_ = limits_.time + limits_.increment;
     }
     moves_left_ = limits_.moves;
 }
@@ -32,13 +30,13 @@ bool TimeControl::updateTime(const int64_t elapsed_millis) noexcept {
     if (limits_.moves > 0) {
         if (moves_left_ == 1) {
             moves_left_ = limits_.moves;
-            time_left_ += limits_.time;
+            time_left_ += limits_.time + limits_.increment;
         } else {
             moves_left_--;
         }
     }
 
-    if (limits_.fixed_time == 0 && limits_.time == 0) {
+    if (limits_.fixed_time == 0 && limits_.time == 0 && limits_.increment == 0) {
         return true;
     }
 
--- a/app/src/time/timecontrol.cpp
+++ b/app/src/time/timecontrol.cpp
@@ -17,7 +17,7 @@ TimeControl::TimeControl(const Limits &limits) : limits_(limits) {
     if (limits_.fixed_time != 0) {
         time_left_ = limits_.fixed_time;
     } else if (limits_.time != 0) {
-        time_left_ = limits_.time;
+        time_left_ = limits_.time + limits_.increment;
     } else {
         time_left_ = 0;
     }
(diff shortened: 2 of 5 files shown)


app/src/matchmaking/player.hpp
```diff
@@ -14,9 +14,7 @@ class Player {
     // The timeout threshold for the read engine command.
     // This has nothing to do with the time control itself.
     [[nodiscard]] std::chrono::milliseconds getTimeoutThreshold() const {
-        if (engine.getConfig().limit.nodes != 0     //
-            || engine.getConfig().limit.plies != 0  //
-            || time_control_.isFixedTime()) {
```
Disservin, line 19: mh why remove this?

app/src/time/timecontrol.cpp
```diff
@@ -32,13 +30,13 @@ bool TimeControl::updateTime(const int64_t elapsed_millis) noexcept {
     if (limits_.moves > 0) {
         if (moves_left_ == 1) {
             moves_left_ = limits_.moves;
-            time_left_ += limits_.time;
+            time_left_ += limits_.time + limits_.increment;
```
Disservin, line 33: you sure increments gets added again before the move is played once the timecontrol repeats? i dont think so

app/src/time/timecontrol.cpp
```diff
@@ -65,7 +63,7 @@ std::ostream &operator<<(std::ostream &os, const TimeControl &tc) {
         return os;
     }
 
-    if (tc.limits_.moves == 0 && tc.limits_.time == 0 && tc.limits_.increment == 0) {
+    if (tc.limits_.moves == 0 && tc.limits_.time + tc.limits_.increment == 0) {
```
Disservin, line 66: this change isnt needed, time aint negative right ? and so we have 0 + 0 == 0 and before == 0 && .. == 0

app/src/matchmaking/player.hpp
gahtan-syarif: we have a time control for fixed time, that conditional is to check if there is a time control set

app/src/time/timecontrol.cpp
gahtan-syarif: hmm right, the increment is already added before the movesleft check, ill remove it

app/src/time/timecontrol.cpp
gahtan-syarif: yeah, this ones just a formatting change to be consistent but ill switch it back if you want

gahtan-syarif: switched it back

app/src/time/timecontrol.cpp
gahtan-syarif: removed

vondele:
If I start a match with cutechess at tc 1+0.1, the first 'go' command gives:
go wtime 1000 btime 1000 winc 100 binc 100
so here the engine has exactly 1s for the first move. That seems to match current fast-chess behavior?

Cutechess also seems to reject 0+0.1 as a TC, probably because the time for the first move is 0s?

gahtan-syarif:
> If I start a match with cutechess at tc 1+0.1, the first 'go' command gives: go wtime 1000 btime 1000 winc 100 binc 100 so here the engine has exactly 1s for the first move. That seems to match current fast-chess behavior?
> 
> Cutechess also seems to reject 0+0.1 as a TC, probably because the time for the first move is 0s?

are you sure about this? this is what i get with tc=5+40
Screenshot 2024-07-23 045152

gahtan-syarif:
theres something off here

gahtan-syarif:
hmm idk, closing this pr until i figure out whats going on

gahtan-syarif:
@vondele this is the first command with tc=1+5
`go wtime 6000 btime 6000 winc 5000 binc 5000`
perhaps youre using an older version of cutechess?

gahtan-syarif:
this was added to cutechess 4 years ago https://github.com/cutechess/cutechess/commit/7a4cf236551c74158e7653e0d94a450081f0440a

vondele:
right, I did a test with an old version (1.1.0). Thanks for checking.

gahtan-syarif:
the version on fishtest is cutechess 1.2.0 which already has this implemented, so i think it would be good to update the fast-chess in the fishtest PR to this commit once this gets merged

Disservin:
If cutechess has this i think it's good to also have it, the clocks in my chess club don't seem to do this however, but I don't think we are playing under fide rule.. idk just talking to myself

robertnurnberg:
There is a choice: adding the increment after the move is down to Fisher, iirc. See https://en.wikipedia.org/wiki/Time_control#Chess

There seems to be also Bronstein delay etc, so in theory one could offer these options to the user? Might be overkill to do all, but maybe Fide and Fisher?

AlexBrunetti:
> the clocks in my chess club don't seem to do this however

The clocks in your chess club should indeed have this functionality; it’s a matter of proper setup. Instead of setting the time to (for example) 3:00, you can set it to 3:02. Alternatively, you can set it to 3:00, and when Black (or the arbiter) starts the clock, they should perform a quick "double tap" to add the initial increments to both players' times.

It's worth noting that we're discussing an increment that is typically between 1/50th and 1/100th of the main time, which is unlikely to be decisive, especially if you’re playing a large number of games.

Alex

Disservin:
> There is a choice: adding the increment after the move is down to Fisher, iirc. See https://en.wikipedia.org/wiki/Time_control#Chess

just for documentation this is what we also do

> a specified amount of time is added to the player's main time after each move, unless the player's main time ran out before they completed their move.

Issue #613, opened 2024-07-20
Multipart round identifiers


gahtan-syarif:
in the pgn standard, you can have an extra identifier for subrounds in the round tag. so for example:
`[Round "1.1"]` for when engine A is playing as white
`[Round "1.2"]` for when engine A is playing as black in the same opening
and then `[Round "2.1"]` and so forth

i can make a PR for this if you want? would make it easier to filter and sort games in the pgn @Disservin

PR #639, opened 2024-07-23
feat: add subround identifiers to pgn

gahtan-syarif:
fix https://github.com/Disservin/fast-chess/issues/613


Disservin:
mh id skip this for now and revisit when a need arises, the pgn doc even says it's a bit unusual
"Some organizers employ unusual round designations"

gahtan-syarif:
well theres also an active pr on cutechess side to add this, so ig when they merge it can be the time when we merge this

PR #646, opened 2024-07-24
refactor: better ctrl c behaviour and engine destruction

Disservin:


Disservin:
@vondele if you still have access to a high thread count machine could you give this a try? i had to put some atomics in place and dont want to run into timeout issues with high thread counts

vondele:
let me have a look.

vondele:
I think this branch is doing fine. I can run with concurrency 250 and TC 1+0.01 with both main and this branch (using SF16 as engine). All fine without timeouts. 

The `time` command returns:

Master
real	0m33.238s
user	105m11.953s
sys	5m41.397s

Stop
real	0m33.250s
user	104m45.573s
sys	5m34.776s

so identical as far as I can see. For such a run, in both cases under `top` is see fast-chess using up to 1000% CPU.

vondele:
However, I had to increase 
```
diff --git a/app/src/engine/uci_engine.hpp b/app/src/engine/uci_engine.hpp
index 08d89b5..3503fbc 100644
--- a/app/src/engine/uci_engine.hpp
+++ b/app/src/engine/uci_engine.hpp
@@ -91,7 +91,7 @@ class UciEngine : protected process::Process {
     [[nodiscard]] const EngineConfiguration &getConfig() const noexcept { return config_; }
 
     // @TODO: expose this to the user
-    static constexpr std::chrono::milliseconds startup_time_    = std::chrono::seconds(5);
+    static constexpr std::chrono::milliseconds startup_time_    = std::chrono::seconds(15);
     static constexpr std::chrono::milliseconds ucinewgame_time_ = std::chrono::seconds(60);
     static constexpr std::chrono::milliseconds ping_time_       = std::chrono::seconds(60);
```
To have clean startup.

gahtan-syarif:
i also seem to be getting issues with the program occasionally being stuck before sending uci when the startup_time_ is 5 seconds

Disservin:
Yeah i increased it could be that during high the spawn process takes a bit longer

vondele:
> Yeah i increased it could be that during high the spawn process takes a bit longer

I wonder if one should do something more advanced, like e.g. use a semaphore: Spawn at most 16 games (2x engines), and only spawn 'the next pair' once 2 engines have reported back with uciok. Spawning like 100s of engines on massively multicore CPUs will take time, given the binaries are 100MB in size. The current simpler choice does work however, so that's not super urgent.

Disservin:
> > Yeah i increased it could be that during high the spawn process takes a bit longer
> 
> I wonder if one should do something more advanced, like e.g. use a semaphore: Spawn at most 16 games (2x engines), and only spawn 'the next pair' once 2 engines have reported back with uciok. Spawning like 100s of engines on massively multicore CPUs will take time, given the binaries are 100MB in size. The current simpler choice does work however, so that's not super urgent.

i think this should implement this? c++ has `std::counting_semaphore` but sadly not in c++17

https://github.com/Disservin/fast-chess/compare/master...counting-semaphore

Disservin:
@vondele did you try that branch together with a lower startup time again maybe?

vondele:
not yet, but I can.

vondele:
works well with 25, 5, and 3s startup time, but failed with 1s. I wouldn't push it and go for 5s.

PR #654, opened 2024-07-26
allow higher concurrency with -force

robertnurnberg:
Might need a second pair of eyes, and some testing.

Locally all seems fine.


Disservin:
mh would need to be added to the config, but while this is the easiest way to implement it I dont like that it is a generic option and not like tied to the concurrency option.
`-concurrency 1 force=true` or `-concurrency force=true 1` is I think better but this way of parsing isnt implemented right now

robertnurnberg:
Ok, I guess I leave that to you then. :)

Feel free to close this one, or make changes on top.

If you go for more fancy version, then you may also want to add it to man.

PR #656, opened 2024-07-26
refactor: drop hyphen in name

Disservin:


Disservin:
supersedes https://github.com/Disservin/fast-chess/pull/655 for now

gahtan-syarif:
I think it would be better if the executable stays fast-chess

robertnurnberg:
Well, if we drop the hyphen we should drop it everywhere, including in the binary.

PR #660, opened 2024-07-28
fix: pgn timecontrol formatting for 0 initial time

gahtan-syarif:
fix case where `tc=0+inc` is displayed as `[TimeContol "+inc"]` in pgn


Disservin:
can you add a test to the pgn tests pelase?

gahtan-syarif:
> can you add a test to the pgn tests pelase?

unit test added

PR #663, opened 2024-08-02
chore: update man

gahtan-syarif:


Disservin:
can you update the included cpp version as well?

gahtan-syarif:
> can you update the included cpp version as well?

Ill do it in a seperate pr since i also have another active pr with man changes

PR #666, opened 2024-08-02
refactor: more concise warning message

gahtan-syarif:


Disservin:
this loses information, about it being the last one?

gahtan-syarif:
> this loses information, about it being the last one?

Updated, also removed the score message since the conditional does not check for score

Disservin:
> since the conditional does not check for score

but it does... https://github.com/Disservin/fastchess/blob/f8da6c3161aa9d9fa74cbead89281ad3497165c9/app/src/engine/uci_engine.cpp#L296

gahtan-syarif:
ahh didnt see that

gahtan-syarif:
updated

app/src/engine/uci_engine.cpp
```diff
@@ -329,8 +329,7 @@ std::vector<std::string> UciEngine::lastInfo() const {
 
     if (last_info.empty()) {
         Logger::warn<true>(
-            "Warning; No info line found in the last line which includes the "
-            "score from {}",
+            "Warning; Last info string with score (cp) not found from {}",
```
Disservin, line 332: why `(cp)` .. this is wrong `score mate` or `score cp`

gahtan-syarif: right, updated

Disservin: okay but why did you know add eval? there is no eval in the last info string line.. we expect score not eval

gahtan-syarif: oh, i thought theyre the same thing(?) but removed it anyways

PR #671, opened 2024-08-03
feat: add "trust" engine option from cutechess

gahtan-syarif:
reference: https://github.com/cutechess/cutechess/pull/414


gahtan-syarif:
are stockfish mate scores a guaranteed win? if so, can be used in fishtest as well

PR #673, opened 2024-08-11
feat: add trust option from cutechess

gahtan-syarif:
reference: https://github.com/cutechess/cutechess/pull/414

already tested it, works well


gahtan-syarif:
merge conflicts resolved

PR #677, opened 2024-08-17
Refactor: minimize delay

gahtan-syarif:
I dont think its good to have a logger message there, as it creates a couple hundred microsecond delay between fc sending the go command and fc starting the move countdown timer


Disservin:
Then move it before the go

gahtan-syarif:
Done 👍

PR #680, opened 2024-08-23
refactor: use packedformat for the epd book

Disservin:


Disservin:
removed the
```
    truncate(rounds_);
    shrink();
```
calls and measured over valgrind-massfiv, peak memory consumption went from 350MB down to 150MB for the `UHO_Lichess_4852_v1.epd` book

Disservin:
needs a wrapper for the move counters, and needs to pass chess960 if it is chess960 to the encoding and the pgn book should also use this

gahtan-syarif:
it might be worth adding a unit test for this for cases where:
1. pure fens ( no `;` )
2. epds with hmvc and fmvn movecounters
3. epds with only hmvc/only fmvn counters
4. epds with no movecounters

Disservin:
> 1. pure fens ( no `;` )

no `;` ?

PR #681, opened 2024-08-23
refactor: match generator

Disservin:
Saves memory and for very large round numbers by lazily creating the matches after they finish.
Pictures below are from a test tournament with 2 700 000 rounds.


Disservin:
old:

old

Disservin:
new

new

gahtan-syarif:
Very nice 👍

PR #687, opened 2024-08-25
refactor: add draw by timeout/disconnect/stall/illegal move

gahtan-syarif:
make fc consistent with fide and uscf rules as well as with cutechess

fix https://github.com/Disservin/fastchess/issues/410

### DIFF
--- a/app/src/matchmaking/match/match.cpp
+++ b/app/src/matchmaking/match/match.cpp
@@ -338,7 +338,7 @@ void Match::setEngineCrashStatus(Player& us, Player& them) {
 
     data_.termination = MatchTermination::DISCONNECT;
     
-    if (board_.hasSufficientMatingMaterial(~board_.sideToMove())) {
+    if (board_.them(board_.sideToMove()).count() > 1) {
         us.setLost();
         them.setWon();
         data_.reason      = color + Match::DISCONNECT_LOSE_MSG;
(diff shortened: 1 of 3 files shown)


gahtan-syarif:
```
+ grep 'Finished game'
+ sort -n -k 3
+ diff ./cutechess-cli-out.finished ./fastchess-out.finished
35c35
< Finished game 35 (sf_1 vs sf_2): 1/2-1/2 {Draw by adjudication}
---
> Finished game 35 (sf_1 vs sf_2): 1/2-1/2 {Draw by insufficient mating material}
```
how do i access the pgn for this?

gahtan-syarif:
done 👍

gahtan-syarif:
Btw finding all the situations in which only one side has no legal set of moves for checkmate is quite complex, so i decided to take the cutechess route of doing it by adjudicating as draws when the opponent only has a lone king: https://github.com/cutechess/cutechess/pull/717

Disservin:
mh does cutechess have the same termination strings? i find this weird tbh, maybe it should be behind a "strict fide" feature but i don't care too much about that as of now because it's not the main focus

gahtan-syarif:
Yeah cutechess has the same termination strings

Disservin:
mh.. i don't really like this, this primarily to test engines, we don't need 100% fide compliance, if so the first illegal move should be ignored and the engine asked to make another move and only lose on the second move. Making the engine lose in 100% of the cases is more consistent imo

PR #690, opened 2024-08-26
fix: games not continuing properly from last config savepoint

gahtan-syarif:
problem seem to have occured after the matchgenerator refactor


Disservin:
we probably should have a unit test for the match generator and assert that the return value of next matches what we expect it to be

PR #692, opened 2024-08-28
fix: only extract the score from info score cp|mate lines

Disservin:


app/src/engine/uci_engine.cpp
```diff
@@ -292,9 +292,10 @@ void UciEngine::writeLog() const {
 
 std::string UciEngine::lastInfoLine() const {
     // iterate backwards over the output and save the first line
-    // that contains "info", "score" and "multipv 1" if it contains multipv
+    // that contains "info", "score", "cp" | "mate" and "multipv 1" if it contains multipv
```
robertnurnberg, line 295: Comment does not correspond to code?

joergoster:
I must say I don't understand this patch. Why not simply saving the score from the last info line as is? Whether this is from a fail-high or -low doesn't matter for the pgn. But outputting no score or whatever might also confuse some GUIs ... probably needs some testing.

gahtan-syarif:
> I must say I don't understand this patch. Why not simply saving the score from the last info line as is? Whether this is from a fail-high or -low doesn't matter for the pgn. But outputting no score or whatever might also confuse some GUIs ... probably needs some testing.

The score that is parsed isnt just used for the pgn, its also used for the resignation and draw adjudication tracker so parsing upper/lowerbound scores could affect results

PR #696, opened 2024-08-31
refactor: change pgn mate score format from moves to plies

gahtan-syarif:
i think this is the correct conversion?
based on SF's `auto m = (mate.plies > 0 ? (mate.plies + 1) : mate.plies) / 2;`

### DIFF
--- a/app/src/matchmaking/match/match.cpp
+++ b/app/src/matchmaking/match/match.cpp
@@ -53,7 +53,9 @@ void Match::addMoveData(const Player& player, int64_t measured_time_ms, int64_t 
         ss << (move_data.score >= 0 ? '+' : '-');
         ss << std::fixed << std::setprecision(2) << (float(std::abs(move_data.score)) / 100);
     } else if (score_type == engine::ScoreType::MATE) {
-        ss << (move_data.score > 0 ? "+M" : "-M") << std::to_string(std::abs(move_data.score));
+        uint64_t plies = board_.sideToMove() == Chess::Color::WHITE ? std::abs(move_data.score) * 2 
+                                                                    : std::abs(move_data.score) * 2 - 1;
+        ss << (move_data.score > 0 ? "+M" : "-M") << std::to_string(plies);
     } else {
         ss << "ERR";
     }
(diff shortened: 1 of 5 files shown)


gahtan-syarif:
ok, verified it with cutechess and its correct

app/src/matchmaking/match/match.cpp
```diff
@@ -52,7 +52,7 @@ void Match::addMoveData(const Player& player, int64_t measured_time_ms, bool leg
         ss << (move_data.score >= 0 ? '+' : '-');
         ss << std::fixed << std::setprecision(2) << (float(std::abs(move_data.score)) / 100);
     } else if (score_type == engine::ScoreType::MATE) {
-        uint64_t plies = board_.sideToMove() == Chess::Color::WHITE ? std::abs(move_data.score) * 2 
+        uint64_t plies = board_.sideToMove() == chess::Color::WHITE ? std::abs(move_data.score) * 2
```
vondele: `chess::Color::WHITE` doesn't this depend on which side is winning ?

gahtan-syarif: yeah, this one is outdated, this is the new one:
```
uint64_t plies = move_data.score > 0 ? std::abs(move_data.score) * 2 - 1
                                             : std::abs(move_data.score) * 2;
```

app/src/matchmaking/match/match.cpp
```diff
@@ -52,7 +52,9 @@ void Match::addMoveData(const Player& player, int64_t measured_time_ms, bool leg
         ss << (move_data.score >= 0 ? '+' : '-');
         ss << std::fixed << std::setprecision(2) << (float(std::abs(move_data.score)) / 100);
     } else if (score_type == engine::ScoreType::MATE) {
-        ss << (move_data.score > 0 ? "+M" : "-M") << std::to_string(std::abs(move_data.score));
+        uint64_t plies = move_data.score > 0 ? std::abs(move_data.score) * 2 - 1
```
robertnurnberg, line 55: Abs not needed here I think. Similarly below.

gahtan-syarif: Right, its always positive, fixed 👍

PR #697, opened 2024-08-31
fix: missing newline between printElo and printSPRT

gahtan-syarif:
`Ptnml(0-2): [0, 0, 1, 1, 0], WL/DD Ratio: infLLR: 0.02 (-2.94, 2.94) [0.00, 5.00]`


Disservin:
whops

PR #701, opened 2024-09-02
feat: add option to only log stderr

gahtan-syarif:
this acts as replacement of cutechess `stderr=arg`


Disservin:
Mh tbh I dont think this is really meaningful, you will only get the stderr of the engine which is likely not much and no clear idea how to reproduce.
Also stderr is currently not tracked on windows.

gahtan-syarif:
The goal is mostly as a replacement to cutechess 'stderr=arg',  i think theres some value in it as regular log files can get insanely big on long tests and sometimes you just want to see if something can run without any errors

Disservin:
well but the usage is different, stderr got added because cutechess completely lacked that functionality and one couldnt see the stderr of the engines and the way they have it implemented right now was just the easiest route https://github.com/cutechess/cutechess/issues/8

Disservin:
i could imagine a reduced log, which only prints engine outputs

Disservin:
ill close this until then

PR #709, opened 2024-09-25
refactor: simplify move legality check

gahtan-syarif:
reduce calls to setEngineIllegalMoveStatus function

### DIFF
--- a/app/src/matchmaking/match/match.cpp
+++ b/app/src/matchmaking/match/match.cpp
@@ -249,7 +249,7 @@ bool Match::playMove(Player& us, Player& them) {
     auto best_move = us.engine.bestmove();
 
     const auto move   = best_move ? uci::uciToMove(board_, *best_move) : Move::NO_MOVE;
-    const auto is_uci = isUciMove(move);
+    const auto is_uci = isUciMove(best_move.value());
     const auto legal  = isLegal(move) && is_uci;
 
     const auto timeout  = !us.updateTime(elapsed_millis);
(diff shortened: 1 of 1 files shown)


app/src/matchmaking/match/match.cpp
```diff
@@ -248,13 +248,9 @@ bool Match::playMove(Player& us, Player& them) {
...
 
-    if (best_move && !isUciMove(best_move.value())) {
-        setEngineIllegalMoveStatus(us, them, best_move, true);
-        return false;
-    }
-
-    const auto move  = best_move ? uci::uciToMove(board_, *best_move) : Move::NO_MOVE;
-    const auto legal = isLegal(move);
+    const auto move   = best_move ? uci::uciToMove(board_, *best_move) : Move::NO_MOVE;
+    const auto is_uci = isUciMove(best_move.value());
```
Disservin, line 252: This will throw an exception if no bestmove is found just use value_or("")

Disservin (changes requested):

app/src/matchmaking/match/match.cpp
gahtan-syarif: alright, fixed 👍

Disservin:
Or mh.. then the warning is wrong..

gahtan-syarif:
which warning?

gahtan-syarif:
if ur talking about this i think it can just be removed https://github.com/Disservin/fastchess/blob/865610cbfc2965a5711bbe85a4ffa715d597d877/app/src/matchmaking/match/match.cpp#L410
i dont see the point of the warning since the move is adjudicated as illegal anyway, im sure the user can figure out why the move is illegal

Disservin:
mh id rather prefer that since there's no explicit definition of move format really

PR #715, opened 2024-09-27
test: add tests for timecontrol

gahtan-syarif:


app/tests/timecontrol_test.cpp
```diff
@@ -0,0 +1,59 @@
...
+        TimeControl::Limits limits;
+        limits.moves = 3;
+        limits.time = 10000;
+        limits.increment = 100;
+        limits.timemargin = 100;
+
+        TimeControl tc(limits);
+
+        CHECK(tc.updateTime(5555) == true);
+        CHECK(tc.getTimeLeft() == 4645);
```
Disservin, line 18: instead of hardcoding the value here you can instead write the expression which calculates this, like the time + increment - 5555 + increment, makes it easier to follow

gahtan-syarif: ahh ok

gahtan-syarif: fixed 👍

PR #723, opened 2024-09-30
fix: dont add movedata if we have a timeout

Disservin:


gahtan-syarif:
@Disservin maybe its better to add move data if its not <none> instead? Since knowing how much time the engine takes to move when it output bm during a timeloss can be beneficial

Disservin:
changed good idea

PR #730, opened 2024-10-01
refactor: restrict tournament format to roundrobin

gahtan-syarif:
doing this since gauntlet and other formats aren't supported yet


Disservin:
oh hol up

PR #734, opened 2024-10-02
fix: game_id and pair_id gets reset to 1 after loading from config

gahtan-syarif:
i think this fixes it


Disservin:
you should rather fix this in the match generator

gahtan-syarif:
ok fixed i think

PR #753, opened 2025-01-04
refactor: improve memory usage

Disservin:


Disservin:
before:
grafik

Disservin:
now:

grafik

Disservin:
ultimately it is probably better to just read from the file directly

PR #765, opened 2025-01-10
fix: dir no longer being able to use for path creation

Disservin:
fixes https://github.com/Disservin/fastchess/issues/764


Disservin:
not tested

PR #773, opened 2025-01-12
fix: missing restart option in json config file

gahtan-syarif:


gahtan-syarif:
hmm im not sure which test failed

PR #782, opened 2025-01-17
avoid informal contractions in error messages

robertnurnberg:
As fastchess matures, maybe so should its error messages. ;)

### DIFF
--- a/app/src/engine/uci_engine.cpp
+++ b/app/src/engine/uci_engine.cpp
@@ -80,7 +80,7 @@ process::Status UciEngine::isready(std::chrono::milliseconds threshold) {
     }
 
     if (res != process::Status::OK) {
-        Logger::trace<true>("Engine {} didn't respond to isready.", config_.name);
+        Logger::trace<true>("Engine {} did not respond to isready.", config_.name);
         Logger::warn<true>("Warning; Engine {} is not responsive.", config_.name);
 
         return res;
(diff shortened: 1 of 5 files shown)


app/src/engine/uci_engine.cpp
```diff
@@ -254,7 +254,7 @@ void UciEngine::sendSetoption(const std::string &name, const std::string &value)
     auto option = uci_options_.getOption(name);
 
     if (!option.has_value()) {
-        Logger::info<true>("Warning; {} doesn't have option {}", config_.name, name);
+        Logger::info<true>("Warning; {} does not have option {}", config_.name, name);
```
Disservin, line 257: fishtest reads this line and checks for "doesn't have option" so that would also need a change

app/tests/doctest/doctest.hpp
```diff
@@ -7103,4 +7103,4 @@ DOCTEST_SUPPRESS_COMMON_WARNINGS_POP
 #ifdef DOCTEST_UNDEF_NOMINMAX
 #undef NOMINMAX
 #undef DOCTEST_UNDEF_NOMINMAX
-#endif // DOCTEST_UNDEF_NOMINMAX
\ No newline at end of file
+#endif // DOCTEST_UNDEF_NOMINMAX
```
Disservin, line 7106: dont change this file

Disservin: it's a third party lib in case you didnt notice, it should ideally go into third_party directory actually

Disservin:
there is no real pressure to open a pr change to fishtest right after this is merged, should just be kept in mind..

robertnurnberg:
I am sorry. Looking at the "switch to fastchess" PR on fishtest I realize that cutechess uses the same contractions in its error messages. So my PR missed leaving them as they are for the cutechess compatibilty mode. It is probably best to simply revert this commit. Sorry again.

Disservin:
ah forgot about that point as well haha all good

PR #805, opened 2025-02-22
feat: add crc32 checksum verification pgn

Disservin:
Add `-crc32 pgn=true;` to the options. If the pgn file already exists then the starting crc32 is calculated.
It should produce a line like this at the end just before exiting.
```
Games written to test.pgn with CRC32: 0x182d613c
```

where the crc32 should match the following in python
```python
import zlib
print(hex(zlib.crc32(b'mydata')))
```


robertnurnberg:
Very nice new feature! Tiny nit:

```
Games written to debug.pgn with CRC32: 0x25eee6a6
Finished match
```

Could we (a) change the order of the two statements, and (b) always write the part without CRC32 if that option is not true, and pgn output was requested?

PR #817, opened 2025-03-15
Add Windows support for CPU affinity for >64 CPUs.

JoostHouben:
On supported Windows platforms (Win11 or Server 2022), use newer APIs to set CPU affinity correctly when there are more than 64 logical CPUs.

Resolves Disservin/fastchess#816, at least for those newer Windows platforms.

With this change, running concurrency 64 on a 7980X (64 cores, 128 SMT threads) on Windows 11 and enabling CPU affinity now results in the expected CPU utilization:
Screenshot 2025-03-15 155325

As you can see, the first logical CPU of each of the 64 physical cores is now being utilized. This results in better and more consistent performance.

I've also validated that the app still works on Windows 10 (where `SetProcessDefaultCpuSetMasks` is not available), both with and without setting CPU affinity. Note that the Windows 10 system is a 4 core / 8 thread machine. I don't think testing on a >64 thread Windows 10 system is required.

### DIFF
--- a/app/src/affinity/affinity.hpp
+++ b/app/src/affinity/affinity.hpp
@@ -3,6 +3,7 @@
 #include <core/logger/logger.hpp>
 
 #ifdef _WIN64
+#    include <processthreadsapi.h>
 #    include <windows.h>
 #elif defined(__APPLE__)
 #    include <mach/thread_act.h>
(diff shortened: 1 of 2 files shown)


JoostHouben:
Note I built the Windows binary using:
```
msys2_shell -defterm -here -no-start -ucrt64 -c "make -j build=release"
```

And the toolchain was installed from the msys2 shell using:
```
pacman -S mingw-w64-ucrt-x86_64-gcc
```

I'm not sure if the current Github workflow is configured correctly to use the UCRT. If building with a toolchain targeting an older Windows C runtime, the changes may fail to compile due to `SetProcessDefaultCpuSetMasks` not being defined. So the Github workflow may need to be updated.

If a maintainer approves the workflow run, I guess we will see.

app/src/affinity/affinity.hpp
```diff
@@ -21,11 +22,43 @@ namespace affinity {
...
 
     for (const auto& cpu : cpus) {
-        assert(cpu <= 63);
-
+        if (cpu > 63) {
+            Logger::print<Logger::Level::ERR>(
+                "Setting affinity for more than 64 logical CPUs is not supported: requires at least Windows 11 or "
+                "Windows Server 2022.");
+            return false;
+        }
```
Disservin, line 61: this is redudant I think, there already is a check for this earlier during config checking, see: https://github.com/Disservin/fastchess/blob/master/app/src/cli/sanitize.cpp#L42

Disservin:
Wow! Thanks a lot for diagnosing this and fixing it :D

app/src/affinity/affinity.hpp
JoostHouben: Hmm, I don't think so. The check you linked is for concurrency > 63, but the check here is for cpu > 63. Those aren't equivalent.

The [affinity manager](https://github.com/Disservin/fastchess/blob/master/app/src/affinity/affinity_manager.hpp) is designed to initially only use one logical SMT CPU per physical core, and only use the second logical SMT CPU per physical core when it runs out. This is done by first assigning only even logical CPU idxs, and then assigning odd logical CPU idxs when it runs out. So if you have more than 32 physical cores and have SMT (resulting in >64 logical CPUs), the affinity manager will assign logical CPU idxs >63 as soon as you request a concurrency of more than 32.

So in particular, under the following conditions the check in sanitize.cpp will pass but the check in affinity.hpp will be triggered:

 - You're on Windows 10 or earlier.
 - You request a concurrency of more than 32 but less than 64.
 - You have more than 64 logical CPUs (for example because you have more than 32 physical cores and have SMT).

In testing I can replicate this by temporary disabling the new Windows 11 API path I added, and then requesting a concurrency of 33. This triggers the warning I added.

As an aside, I'm not sure what the origin of the check in sanitize.cpp is. My guess is that it's because Windows 10 and before are unable to schedule the threads within a single process across more than 64 CPUs. Therefore, if you request a concurrency of >63 (+ the main thread), the threads internal to fastchess will have some CPU contention, even if the engine processes themselves are possibly still correctly scheduled across >64 CPUs (Windows 10 should still be capable of putting separate processes in separate CPU groups to accomplish this). Assuming that's true, I'm not entirely sure it's a reasonable restriction. After all, the threads internal to fastchess should be largely I/O bound, not CPU bound, so CPU contention is not likely to be a major issue. I suppose it's still a reasonable safe-guard against issues if you're being very conservative though.

Regardless, since the sanitize.cpp and affinity.hpp checks are checking different conditions and guarding against different issues, I think it's fine to let them co-exist.

JoostHouben: > the STL threads are limited to 64 threads

I don't think that's true. I was able to launch 128 concurrent threads on my Windows 10 laptop with no issues.

The only limitation is that whatever threads are spawned will share the same processor group of at most 64 CPUs. So I think it would make sense to lift the restriction of concurrency <64, or to at least allow it to be bypassed with `--force-concurrency`.

Anyway, I guess that's outside the scope of this PR.

Disservin:
thanks!

app/src/affinity/affinity.hpp
Disservin: > I don't believe that's true. I was able to launch 128 concurrent std::threads on my Windows 10 laptop with no issues.

Well yeah launching them shouldn't error but they will all end up in the same processor group as you said, which isn't really ideal for the engine processes which I think also end up in the same processor group

JoostHouben: Hmm, I thought the engine processes could still be in a different processor group. But on second thought, you might be right that they inherit the processor group from the parent process.

PR #818, opened 2025-03-16
Add Syzygy TB adjudication

JoostHouben:
Add W/D/L adjudication using Syzygy tablebases. Resolves Disservin/fastchess#593.

Tablebase adjudication is invoked by adding the CLI argument `-tb` with a semicolon-separated list of TB directories. E.g., `-tb C:\syzygy345;C:\wdl6`. This matches the CLI of cutechess. Though note that additional options supported by cutechess (` -tbpieces N` and `-tbignore50`) were not implemented (though this would be straight-forward to add).

This uses https://github.com/AndyGrant/Pyrrhic. Modifications made to the third_party/pyrrhic directory:
 - Renamed `tbprope.c` to `tbprobe.cpp` so that the project's C++ configuration is used.
 - Modified `tbconfig.h` to utilize `chess.hpp` for generating attacks and for bitboard manipulation.

Additionally, `third_party/pyrrhic/tbprobe.cpp` was added to `SRC_FILES` in the app's `Makefile`.

Testing done:

 - On Windows:
    - Builds.
    - Help shows the new argument.
    - Running with `-tb` (and no path) results in the error: `Must provide a ;-separated list of Syzygy tablebase directories.`.
    - Running with `-tb C:\invalid_path` (and valid required arguments) results in the error: `Failed to load Syzygy tablebases from the following directories: C:\invalid_path`.
    - Ran a 200 game tournament without TB adjudication and verified:
       - Tournament completed successfully.
       - TB adjudication config was written to the config.json file.
    - Ran a 200 game tournament _with_ TB adjudication (6-men), and verified:
       - Tournament completed successfully.
       - TB adjudication config was written to the config.json file.
       - Syzygy TB adjudication messages appear in the terminal output (`{White wins by Syzygy TB adjudication}`, `{Black wins by Syzygy TB adjudication}`, `{Draw by Syzgy TB adjudication}`).
       - Verified that the same adjudication messages appear in the pgn output file, as well as `[Termination "adjudication"]`.
       - Spot-checked TB-adjudicated games from the pgn output file (a win by white after white moved, a win by white after black moved, a win by black after white moved, a win by black after black moved, a draw after white moved, a draw after black moved), and verified:
           - The end state is a 6-men position.
           - The adjudication message matches the game result.
           - The adjudication message / game result matches the Syzygy result shown on https://syzygy-tables.info/ for that position.
 - On Linux:
    - Builds.
    - Help shows the new argument.
[...]

PR #819, opened 2025-03-22
A few enhancements to the TB adjudication feature

JoostHouben:
- Add support for cutechess options `-tbpieces` and `-tbignore50`.
 - Tweak adjudication wording to be identical to cutechess.
 - Add TB adjudication to the cutechess compatibility test, as suggested by @robertnurnberg on #818.

The modified cutechess compatibility test will look for 345-men TBs and use them if present. Though because of the `-tbpieces 4` option only the 34-men TBs are actually used for adjudication. The modified workflow will download the 345-men TBs for the test to use, and will cache the files if possible.

The `-tbignore50` option can't be tested via the compatibility test because cursed wins/losses are too rare an occurrence. A new unit test is added instead, and a single 5-men WDL file was added to support this (~3.4 MB).


robertnurnberg:
Very nice. This could be of interest to @gahtan-syarif as well.

What I do not quite understand yet is why the tb files have to be uploaded to this repo as well. Is this to allow also local testing, outside of CI?

JoostHouben:
> What I do not quite understand yet is why the tb files have to be uploaded to this repo as well. Is this to allow also local testing, outside of CI?

Yes, it's to make the unit tests independent of the execution environment / github workflows. Otherwise a developer would have to manually download TBs in order to run unit tests locally, which would be an unpleasant experience.

.github/workflows/cutechess_tests.yml
```diff
@@ -28,6 +28,27 @@ jobs:
           clang++ --version
           g++ --version
 
+      - name: Create cutechess-testing directory
+        run: |
+          mkdir cutechess-testing
+
+      - name: cache syzygy
+        id: cache-syzygy
+        uses: actions/cache@v4
```
Disservin, line 37: can you use the commit sha of version 4 please? slightly more secure and best practices

JoostHouben: Sure, I can do that. But isn't using a major version tag better because it allows us to automatically pick up security fixes?

gahtan-syarif:
Nice addition :)


For the change in the adjudication message:
```cpp
     inline static constexpr char ADJUDICATION_TB_WIN_MSG[]  = " wins by adjudication: SyzygyTB";
```
Is it the same as the adjudication message in cutechess? Since ideally it would be the same for both cutechess and fastchess, or at least when the cutechess output format is chosen

.github/workflows/cutechess_tests.yml
Disservin: the actions were written when i didn't really know about the sha commit usage for actions, but github recommends that nowadays https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions
"Pin actions to a full length commit SHA"

i don't even know if github updates old tags for security fixes, i can update those tags in the future too myself

JoostHouben:
@gahtan-syarif

> Is it the same as the adjudication message in cutechess? Since ideally it would be the same for both cutechess and fastchess, or at least when the cutechess output format is chosen

Yes, see PR description:

> Tweak adjudication wording to be identical to cutechess.

The updated cutechess compatibility test also enforces that the message is identical to what's used in cutechess.

gahtan-syarif:
Ahh ok didnt see that, thank you!

.github/workflows/cutechess_tests.yml
JoostHouben: Ok, pushed an update that uses the SHA hash of the most recently tagged release.

> i don't even know if github updates old tags for security fixes

Actually yes! The v4 tag points to v4.2.3 released 3 days ago, see: https://github.com/actions/cache/tags

PR #820, opened 2025-03-23
Various QoL features and fixes

JoostHouben:
A bunch of small quality-of-life features and fixes:
 - Print a message if the SPRT has completed, and indicate whether H0 or H1 was accepted. Previously, the match would just end with no clear indication of why (unless you had logging enabled and reviewed the log). Ex.:
    ```
    SPRT ([0.00, 100.00]) completed - H0 was accepted
    ```
 - Print diagnostic messages for the following abnormal terminations: signal received, engine not responsive. Previously, these were only log messages, now they also show in the terminal. Ex.:
    ```
    Received signal, stopping tournament.
    ```
 - In case of an abnormal termination (such as SIGINT or an engine crashing without `-recover`): instead of printing 'finished match', print a message indicating that the tournament was interrupted, and automatically generate the command needed to resume the tournament if desired. Also set exit code to 1. Ex.:
    ```
    Tournament was interrupted. To resume the tournament, run: C:\Git\fastchess\fastchess.exe -config 
    file=C:\Chess\fastchessConfig.json
    ```
    Or:
    ```
    Tournament was interrupted. To resume the tournament, run: ./fastchess -config file=config.json
    ```
 - If the tournament is forcefully stopped by fastchess (by setting `atomic::stop` to `true`), don't mark engines as stalled/disconnected. (Note that in such a case `Process::readOutput` always returns an error state, so the match code will think the engine has stalled/disconnected even if the process is stilly healthy; we don't want to surface that as a crash to the user.)
 - Print the percentage of the LLR relative to the SPRT bounds, similar to cutechess. Ex.:
    ```
    LLR: -0.30 (-10.2%) (-2.94, 2.94) [0.00, 100.00]
    ```
 - Fix return value of `SPRT::getLLR` (H0 vs H1 didn't matter previously as it was only compared to CONTINUE).

More complete example for SPRT, to show how some of it fits together:
```
Results of sf_1 vs sf_2 (6 plies - 8 plies, 1t, 16MB, UHO_4060_v2.epd):
Elo: -398.19 +/- 82.31, nElo: -591.37 +/- 54.17
LOS: 0.00 %, DrawRatio: 7.59 %, PairsRatio: 0.01
Games: 158, Wins: 5, Losses: 134, Draws: 19, Points: 14.5 (9.18 %)
Ptnml(0-2): [58, 14, 6, 1, 0], WL/DD Ratio: 2.00
LLR: -2.97 (-100.8%) (-2.94, 2.94) [0.00, 10.00]
SPRT ([0.00, 10.00]) completed - H0 was accepted
Finished match
```

Associated refactors:
 - `atomic::stop` (and the new `atomic::abnormal_termination`) is now declared in globals.hpp instead of being declared ad-hoc at various sites of usage.
 - The config file path is now defaulted to config.json directly in `config::Tournament`, instead of doing that in `BaseTournament`. This allows other code (such as the diagnostic in main.cpp) to reference `config::Tournament::config_name` directly without having to worry about defaulting it if it's empty.

### DIFF
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -34,8 +34,6 @@ int main(int argc, char const* argv[]) {
             } else {
                 Logger::print("Tournament was interrupted.");
             }
-        } else {
-            Logger::print("Finished match");
         }
     } catch (const std::exception& e) {
         stopProcesses();
@@ -46,6 +44,8 @@ int main(int argc, char const* argv[]) {
     }
 
     stopProcesses();
+
+    Logger::print("Finished match");
 
     return atomic::abnormal_termination ? EXIT_FAILURE : EXIT_SUCCESS;
 }
--- a/app/src/main.cpp
+++ b/app/src/main.cpp
@@ -26,6 +26,17 @@ int main(int argc, char const* argv[]) {
     try {
         auto tournament = TournamentManager();
         tournament.start(cli::Args(argc, argv));
+
+        if (atomic::abnormal_termination) {
+            if (argc > 0 && config::TournamentConfig) {
+                Logger::print("Tournament was interrupted. To resume the tournament, run: {} -config file={}", argv[0],
+                              config::TournamentConfig->config_name);
+            } else {
+                Logger::print("Tournament was interrupted.");
+            }
+        } else {
+            Logger::print("Finished match");
+        }
     } catch (const std::exception& e) {
         stopProcesses();
 
@@ -36,7 +47,5 @@ int main(int argc, char const* argv[]) {
 
     stopProcesses();
 
-    Logger::print("Finished match");
-
-    return 0;
+    return atomic::abnormal_termination ? EXIT_FAILURE : EXIT_SUCCESS;
 }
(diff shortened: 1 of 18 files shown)


app/src/main.cpp
```diff
@@ -26,6 +26,17 @@ int main(int argc, char const* argv[]) {
...
+
+        if (atomic::abnormal_termination) {
+            if (argc > 0 && config::TournamentConfig) {
+                Logger::print("Tournament was interrupted. To resume the tournament, run: {} -config file={}", argv[0],
+                              config::TournamentConfig->config_name);
+            } else {
+                Logger::print("Tournament was interrupted.");
+            }
+        } else {
+            Logger::print("Finished match");
```
Disservin, line 38: I think we always need to print "Finished match", since fishtest tries to match this output line, https://github.com/official-stockfish/fishtest/blob/master/worker/games.py#L1003

JoostHouben: Thanks for pointing that out. Restored the old behavior of always printing 'Finished match'. We still print the interruption message in case of abnormal termination though.

Disservin:
Thanks! Especially that fix for the engine status :D

PR #829, opened 2025-04-12
add support for gzipped books

robertnurnberg:
This allows to specify the book names `foo.epd.gz` or `bar.pgn.gz`, and the files will be decompressed on the fly.

I find this feature very convenient, and it may allow to streamline the book handling for Fishtest in future.

If this is merged, some checks may need to be added to the CI to see that this works on all architectures (it should).

Also, it may be time to retire the `ZLIB` makefile flag and always compile with zlib support.

### DIFF
--- a/app/src/game/book/epd_reader.cpp
+++ b/app/src/game/book/epd_reader.cpp
@@ -7,6 +7,10 @@
 
 #include <core/memory/heap_str.hpp>
 
+#ifdef USE_ZLIB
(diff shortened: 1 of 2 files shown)


Disservin:
main reason for the zlib flag is that a) i think zlib is not by default available and b) licensing

robertnurnberg:
Right! I had forgotten about the lib itself.

PR #843, opened 2025-05-07
Use "fischerandom" for PGN output in cutechess mode

robertnurnberg:
Not sure if this is the best way to implement this (and if you want to extend the cutechess compatibility behaviour to the PGN output).


Disservin:
i mean it is correct but don't think anyone needs this currently?

PR #850, opened 2025-05-31
feat: allow adjudication to be only for won/lost or/and draws

Disservin:
needs some testing


man.md
```diff
@@ -99,6 +99,12 @@ The following options are available:
 - -tbignore50  
     Disable the fifty move rule for tablebase adjudication.
 
+- -tbadjudicate SETTING
+    Control when tablebase adjudication is applied. SETTING can be:
+    - WIN_LOSS: Only adjudicate winning/lost positions
```
robertnurnberg, line 104: Tiny nit: `won/lost`

app/src/matchmaking/match/match.cpp
```diff
@@ -544,30 +544,35 @@ bool Match::adjudicate(Player& us, Player& them) noexcept {
...
     if (config::TournamentConfig->tb_adjudication.enabled && tb_adjudication_tracker_.adjudicatable(board_)) {
         const GameResult result = tb_adjudication_tracker_.adjudicate(board_);
-        if (result != GameResult::NONE) {
-            // Note: 'them' is the player to move at this point, so the result is from their perspective.
+        const auto desired_adju = config::TournamentConfig->tb_adjudication.result_type;
 
-            if (result == GameResult::DRAW) {
-                us.setDraw();
-                them.setDraw();
+        if ((result == GameResult::WIN || result == GameResult::LOSE) &&
```
robertnurnberg, line 549: For consistency LOSE could be renamed to LOSS, though maybe not as part of this pr.

man.md
```diff
@@ -99,6 +99,12 @@ The following options are available:
 - -tbignore50  
     Disable the fifty move rule for tablebase adjudication.
 
+- -tbadjudicate SETTING
+    Control when tablebase adjudication is applied. SETTING can be:
+    - WON_LOST: Only adjudicate winning/lost positions
+    - DRAW: Only adjudicate drawn positions
```
robertnurnberg, line 105: Personally I would go with `- WIN_LOSS: Only adjudicate won/lost positions`. But if you prefer `WON_LOST` then the other should be `DRAWN` I guess?

Disservin:
```
-tbadjudicate SETTING
    Control when tablebase adjudication is applied. SETTING can be:
    - WIN_LOSS: Only adjudicate winning/lost positions
    - DRAW: Only adjudicate drawn positions
    - BOTH: Adjudicate both wins and draws (default)
```

PR #852, opened 2025-06-02
Gauntlet

Matthies:
A lot of code duplication from roundrobin. Maybe better to do the tweaks needed for gauntlet directly in the roundrobin code?


Matthies:
> oh nice :D what changes did you make to Gauntlet.cpp compared to roundrobin ? just get rid of sprt?

Besides removin sprt I remember just two things in the scheduler that I changed:

- Reset player1 when it exceeds the number of seeded engines in next().
- Use number of seeded engines to calculate number of games in total().

I will prepare a second PR with these changes in roundrobin. Probably better.

PR #858, opened 2025-06-14
Fix Makefile

spinojara:
The first error is because the .PHONY line is right before the ifeq. On my system (GNU Make 4.4.1) this means that the ifeq is never executed and there is thus no check for the existence of lowdown. I get the following error when installing:
```
Installing fastchess binary to: /usr/local/bin
/bin/sh: line 1: lowdown: command not found
make[1]: *** [Makefile:45: manpage] Error 127
make: *** [Makefile:37: install] Error 2
```
To fix this, we can move the .PHONY line to e.g. the last line.

The second problem comes with echo. Bash' built-in echo, and GNU echo, both require the -e flag to escape expand. However other echo implementations may be confused by the -e flag. The portable way is to use printf instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,6 @@ MANDIR = $(PREFIX)/share/man/man1
 MANPAGE_SOURCE = man.md  # Path to the markdown file for the man page
 MANPAGE_OUTPUT = fastchess.1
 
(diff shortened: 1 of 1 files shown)


Disservin:
thanks!

PR #862, opened 2025-06-16
Make it possible to put -each argument early

spinojara:
The -each argument does not currently work if it is put before any of the -engine arguments because their configs have not yet been created. This patch makes it so that fastchess always parses the -each argument last.


Disservin:
This seems like a workaround I think there are options which push something into a vector which with this change would happen twice.. rather I'd store the string or what each would do and call that at the end again

PR #864, opened 2025-06-18
Add option to specify cpus with -use-affinity

spinojara:
I have a very specific use case where I need to run fastchess on some very specific cpus. This patch makes it possible to specify an optional argument to -use-affinity as a list of the form "2,4,5-11". If no optional argument is given the behavior is unchanged.


Disservin:
I think you can read minds haha.. I was literally just talking about this with someone else the other day over discord and it was on my todo list 
grafik
though the primary reasoning in our case was the fact that the affinity parser doesn't quite work everywhere #217, we then decided it would also make sense to expose this to the user too :)

Disservin:
looks good with `-use-affinity $(seq -s, 0 2 31)`
grafik

and `seq -s, 0 1 31`

grafik

Disservin:
thanks! btw I think when you generated the new man page, you somehow formatted the man.md file locally there were some additional whitespaces in the generated xxd output so I regenerated that or maybe a bug in lowdown idk, doesn't matter that much :D

spinojara:
Hmm, weird. I have lowdown version 2.0.2.

PR #867, opened 2025-06-21
refactor: also set affinity in manager thread

Disservin:
CPU Affinity Implementation Strategy:

Previously, CPU affinity was set on engine processes after they were already running.

This approach failed because:

1. Chess engines typically spawn worker threads during initialization
2. These worker threads were created before affinity was applied
3. Child threads don't inherit affinity from parent processes retroactively
4. This resulted in engines running on incorrect CPU cores

Current approach:

- Set CPU affinity on the managing thread BEFORE starting any engines
- This ensures all subsequently spawned engine processes inherit the correct affinity
- Engine caching strategy depends on affinity usage:
* With affinity: thread-local engine cache (engines stick to assigned cores)
* Without affinity: global engine cache (engines shared across all threads)

More correct and less timeouts

### DIFF
--- a/app/src/affinity/affinity.hpp
+++ b/app/src/affinity/affinity.hpp
@@ -68,7 +68,7 @@ inline bool setAffinity(const std::vector<int>& cpus, HANDLE process_handle) noe
 
 #elif defined(__APPLE__)
 
-inline bool setAffinity(const std::vector<int>& , pid_t ) noexcept {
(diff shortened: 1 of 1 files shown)


Disservin:
slightly different numbering but much better usage of the individual cores and more ideal other cores

pr:
grafik

old-master:
grafik

Disservin:
fyi @spinojara there were some issues in the past dunno if they affected you (your engine) though

PR #869, opened 2025-06-22
Allow engines to be started from PATH

spinojara:


spinojara:
Ok, I see a problem. Before fastchess could execute a program if it were in your working directory like "-engine cmd=foo". Now we need to do "-engine cmd=./foo". This is what I would expect if I ran fastchess for the first time, but maybe this breaks some configs.

spinojara:
Cutechess only starts engines from path, not from your current directory. So the "opposite" of what fastchess does.

Disservin:
Given the amount of issues I had in #872 I will probably stick with the posix spawn approach and try to add a fallback here which tries to add ./ in front if the execution failed to match the old behavior?

PR #873, opened 2025-06-25
Fix pgn move number error

spinojara:


Disservin:
can you add a test for this ? I assume it's also what cutechess does?

spinojara:
Yes, this is also what cutechess does.

PR #913, opened 2025-09-13
fix: make sure the engine executable exists

leofracca:
Exit right away if the engine isn’t found.
This makes the error message clearer by saying the executable doesn’t exist, rather than showing the previous generic one: "Fatal; <engine_name> engine startup failure: Couldn't start engine process."

### DIFF
--- a/app/src/cli/sanitize.cpp
+++ b/app/src/cli/sanitize.cpp
@@ -133,7 +133,7 @@ void sanitize(std::vector<EngineConfiguration>& configs) {
         }
 
         if (!configs[i].dir.empty() || enginePath.is_absolute()) {
-            if (!(std::filesystem::exists(enginePath) && std::filesystem::is_regular_file(enginePath))) {
+            if (!std::filesystem::is_regular_file(enginePath)) {
                 throw std::runtime_error("Engine binary does not exist: " + enginePath.string());
             }
         }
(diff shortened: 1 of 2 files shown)


Disservin:
This has multiple flaws 
a) the filesystem usage is currently not guarded
b) this assume cmd is an exact path which
might not be the case when used together with dir
c) we had this in the past but removed it because we allowed engines to be started from PATH which means filesystem checks will fail unless the path is expanded which I don't want to do

Disservin:
yeah so for 
a) you'd have to use NO_STD_FILESYSTEM, which currently also used in other places
b) that logic should go in sanitize
c) we can check if it is an absolute path, im not sure how you do the check when dir is set since you might have to check the path..

app/src/cli/sanitize.cpp
```diff
@@ -125,6 +125,19 @@ void sanitize(std::vector<EngineConfiguration>& configs) {
...
         }
+
+#ifndef NO_STD_FILESYSTEM
+        std::filesystem::path enginePath = configs[i].cmd;
+        if (!configs[i].dir.empty()) {
+            enginePath = (std::filesystem::path(configs[i].dir) / configs[i].cmd);
+        }
+
+        if (!configs[i].dir.empty() || enginePath.is_absolute()) {
+            if (!(std::filesystem::exists(enginePath) && std::filesystem::is_regular_file(enginePath))) {
```
Disservin, line 136: is_regular_file implies that the file exists no ? so the previous check is not needed

leofracca: Right, fixed with last commit

PR #914, opened 2025-09-16
fix: Close fds in child processes

MinetaS:


MinetaS:
Since we dup pipe fds to stdio fds with `setup_spawn_file_actions`, all pipe fds should be closed after dup otherwise they are leaked.

"self-pipe trick" thing is removed here; I honestly don't see how engines are supposed to use those pipes for signal processing. Can you explain about its use case, I'm curious about it.

Disservin:
> "self-pipe trick" thing is removed here; I honestly don't see how engines are supposed to use those pipes for signal processing. Can you explain about its use case, I'm curious about it.

Okay so the current implementation uses a poll + read, the timeout on the poll is the remaining thinking time of the engine, meaning if poll times out, the engines time ran out and we return, otherwise we know there is data to be read meaning we can read from the pipe.
If a user specified a long tc like 1h, engines that are about to play the first move get a lot of time, if during that time fastchess receives CTRL + C, the poll will wait for a very long time making the sigint seem unresponsive. In this case the self pipe trick is used, where we write some dummy data so that poll gets notified about the data and we break out of the otherwise blocking poll call.

Meaning in your implementation CTRL + C sent to fastchess will take longer, growing with the engines TC. This is not ideal.

--- 

Does your change actually lead to child processes being killed when fastchess receives SIGKILL, I assume so right ?

--- 

Maybe we don't need the self pipe trick and instead poll every few cycles (TBD) and implement a timeout in some other cheap way.

Disservin:
didn't mean to close

Disservin:
oh i think i might have misunderstood the actual behavior.. the comment (`// keep open for self to pipe trick`) was there because I thought it stops working when there's a close but it seems this is not the case and O_CLOEXEC seems to be a good choice here too, didn't have this in mind

MinetaS:
Oh macOS doesn't have `pipe2`... need to fix that. But I don't own any macOS environments so cannot test it myself.

PR #926, opened 2025-09-30
bugfix: gettid is not available in older glibc versions

AndyGrant:
I understand this to produce identical results; and that gettid() is simply a wrapper for syscall(SYS_gettid) on Linux. I could be mistaken, so do check however you would normally.

### DIFF
--- a/app/src/affinity/affinity.hpp
+++ b/app/src/affinity/affinity.hpp
@@ -12,6 +12,7 @@
 #else
 #    include <sched.h>
 #    include <unistd.h>
+#    include <sys/syscall.h>
 #endif
 
 namespace fastchess {
@@ -168,7 +169,7 @@ template <typename F>
     // dummy
     return 0;
 #    else
-    return gettid();
+    return static_cast<pid_t>(syscall(SYS_gettid));
 #    endif
 }
 #endif


Disservin:
fishtest is compiling from source so for the time being i'll add this..

recently added those static musl builds since i was thinking about moving away for self compiles on fishtest, my message on discord

> been wondering about the way fastchess is used in fishtest and moving away from the fishtest compilation step..
and instead generate the binaries over github CI and downloading them..
a) would allow me to use newer c++ features and other libraries and better build setup
b) would help with compile bugs like https://github.com/Disservin/fastchess/issues/786 which some users experienced

Disservin:
that and the other linked issue from above was actually a compiler bug happening on a specific ubuntu version together with arm which caused a miscompile.. and also cleaning up some parts of the code, like std::filesystem usage which is behind a bunch of flags to check if it is available on the system

Disservin:
the proper build selection is just arch and os, im guessing something like this will work for 99% of users
```python
import platform
import sys

def select_fastchess_asset():
    system = platform.system().lower()
    machine = platform.machine().lower()
    
    if machine in ('amd64', 'x86_64', 'x64'):
        arch = 'x86-64'
    elif machine in ('arm64', 'aarch64'):
        arch = 'arm64'
    elif machine in ('riscv64',):
        arch = 'risc-v'
    else:
        raise ValueError(f"Unsupported architecture: {machine}")
    
    if system == 'linux':
        return f"fastchess-linux-{arch}.tar"
    elif system == 'darwin':
        return f"fastchess-mac-{arch}.tar"
    elif system == 'windows':
        return f"fastchess-windows-{arch}.zip"
    else:
        raise ValueError(f"Unsupported OS: {system}")
```

AndyGrant:
Ah. That looks quite simple. In my mind, I was imagining the effort I did to discern ssse3/avx/avx2/avx512/vnni/sve/dotprod/yadayada.

But you presumably build these to the lowest common denominator.

Issue #898, opened 2025-08-03
Make fastchess exit on warning


spinojara:
For debugging purposes it would be nice to have an option that makes fastchess exit on any warnings and return nonzero.

Issue #886, opened 2025-07-19
Use smarter name defaults


dav1312:
### Engine names

The `-engine` command requires a `name` for each engine. If a name is missing, or if two engines are given the same name, the program throws an error.

1.  If an engine's `name` is not provided, automatically default it to its `cmd` filename (e.g., `path_to_engine\stockfish.exe` becomes `stockfish`).
2.  If duplicate names are detected after defaulting, automatically append a suffix to make them unique, such as `(1)`, `(2)`, etc. For example, two instances of `cmd=stockfish.exe` would become `stockfish (1)` and `stockfish (2)`.

### PGN filenames

The `-pgnout` command requires an explicit `file=filename.pgn` parameter. If it's missing, the program throws an error.

If `-pgnout` is used without a `file` parameter, automatically generate a filename such as: `engine1-engine2_YYYY-MM-DD_HHMMSS.pgn`.

vondele:
As a minor nit, instead of `stockfish (1)` could they be named `stockfish_1`  like in the filename generated for the pgn, this otherwise becomes a difficult and needs escaping (at least on linux).

Issue #857, opened 2025-06-11
Refactor Stats saving and resuming


vondele:
For a match started like:
```
/home/vondele/chess/vondele/workspace/scratch/abcdefgh/testing/fastchess/fastchess \
    -rounds 10 \
    -games 2 \
    -repeat \
    -srand 42 \
    -concurrency 280 \
    --force-concurrency \
    -openings file=/home/vondele/chess/vondele/workspace/data/UHO_Lichess_4852_v1.epd \
              format=epd \
              order=random \
    -ratinginterval 280 \
    -report penta=true \
    -pgnout file=match.pgn \
    -engine name=reference \
            cmd=/home/vondele/chess/vondele/workspace/scratch/abcdefgh/testing/reference/Stockfish/src/stockfish \
    -engine name=step_97fb32d68c61_nn-0861a1e95b35.nnue \
            cmd=/home/vondele/chess/vondele/workspace/scratch/abcdefgh/testing/testing/Stockfish/src/stockfish \
            option.EvalFile=/home/vondele/chess/vondele/workspace/scratch/97fb32d68c61/run/lightning_logs/version_0/checkpoints/nn-0861a1e95b35.nnue \
    -each proto=uci \
          option.Threads=1 \
          option.Hash=16 \
          tc=10+0.1
```
(most importantly `-rounds 10 -games 2 -repeat` and two engines)
The config file contains
```
    "stats": {
        "step_97fb32d68c61_nn-0861a1e95b35.nnue vs reference": {
            "wins": 0,
            "losses": 8,
            "draws": 0,
            "penta_WW": 0,
            "penta_WD": 0,
            "penta_WL": 0,
            "penta_DD": 0,
            "penta_LD": 0,
            "penta_LL": 4
        },
        "reference vs step_97fb32d68c61_nn-0861a1e95b35.nnue": {
            "wins": 11,
            "losses": 0,
            "draws": 1,
            "penta_WW": 5,
            "penta_WD": 1,
            "penta_WL": 0,
            "penta_DD": 0,
            "penta_LD": 0,
            "penta_LL": 0
        }
    }

```
which seems to suggest one engine played 12 games with white and the other only 8. That would be wrong given the options, and is also different from what I see in the pgn:
```
$ grep "\[White" match.pgn  | sort | uniq -c
     10 [White "reference"]
     10 [White "step_97fb32d68c61_nn-0861a1e95b35.nnue"]
```

Disservin:
[...]

PR #927, opened 2025-10-04
perf: minor improvements in parse function

Adi-183:


app/src/cli/cli.hpp
```diff
@@ -134,6 +133,7 @@ class OptionsParser {
 
             try {
                 std::vector<std::string> params;
+                params.reserve(2);
```
Adi-183, line 136: For the -engine option we must specify at least two parameters (cmd and name). Without reserving, pushing these would trigger two allocations (size 0 → 1, then 1 → 2) because GCC’s std::vector typically doubles capacity on growth. Other options like -pgnout or -openings can have more parameters, so we might want to increase the reserved size accordingly.

@Disservin thoughts?

Disservin:
this is really not performance important code, im happy to merge the map -> unordered change but dont think we need to start reserving memory right now

Disservin:
potentially this #886 or #898, something more in depth and substantial would be #857

PR #928, opened 2025-10-06
Update manual

dav1312:
Group options, reorder values in parethesis to display defaults first, add complex example.


Disservin:
the man.md is automatically converted to a manpage using lowdown.

- The double ## generate weird output

grafik

- Weird book highlighting
grafik

PR #1000, opened 2026-04-05
Fix extra initial newline in PGN writing

AndyGrant:
If the initial line is > 80 characters, then a leading newline is added, leading to multiple new lines between the headers and body of the PGN, which breaks the spec., and breaks python-chess at least. PR fixes it. 

This is technically going to be a breaking change for OpenBench. I don't have a great mechanism to force update other than the version value. So if its immaterial to you, bumping the version to 1.8.1 would help me. If not, then I'll go and get some versioning more granular.


Disservin:
i'll bump the version

PR #1003, opened 2026-05-05
refactor: Show Last Extracted Line

Disservin:
@robertnurnberg as discussed


robertnurnberg:
Yeah, looks good. I can prepare a fishtest pr with this update of fastchess and the discussed exceptions for PTs.

PR #1005, opened 2026-05-07
Simplify bestmove treatment

robertnurnberg:
This PR simplifies the logic around the bestmove treatment and aligns the various setEngineStatus functions.


Disservin:
nice!

PR #1006, opened 2026-05-07
Only send warning about missing bestmove if engine did not timeout

robertnurnberg:
No longer send warnings about bestmove extraction if the engine timed out.

I believe this fixes the root cause of those spurious bestmove extraction warnings on fishtest.

### DIFF
--- a/app/src/engine/uci_engine.cpp
+++ b/app/src/engine/uci_engine.cpp
@@ -432,32 +432,42 @@ bool UciEngine::writeEngine(const std::string& input) {
     return process_.writeInput(input + "\n").code == process::Status::OK;
 }
 
-std::optional<std::string> UciEngine::bestmove() const {
+std::optional<std::string> UciEngine::bestmove(bool warn_on_error) const {
+    const auto warn = [&](std::string_view reason, std::string_view last_line = {}) {
+        if (!warn_on_error) {
+            return;
+        }
+
+        if (last_line.empty()) {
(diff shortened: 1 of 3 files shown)


Disservin:
could you update it to this please?
```cpp
std::optional<std::string> UciEngine::bestmove(bool warn_on_error) const {
    const auto warn = [&](std::string_view reason,
                          std::string_view last_line = {}) {
        if (!warn_on_error) {
            return;
        }

        if (last_line.empty()) {
            Logger::print<Logger::Level::WARN>(
                "Warning: No output from {}", config_.name);
        } else {
            Logger::print<Logger::Level::WARN>(
                "Warning: No bestmove found from {} because: {}. Last line was: {}",
                config_.name, reason, last_line);
        }
    };

    const auto lines = getStdoutLines();

    if (lines.empty()) {
        warn("No output");
        return std::nullopt;
    }

    const auto& last_line = lines.back()->line;

    if (last_line.rfind("bestmove", 0) != 0) {
        warn("Line does not start with 'bestmove'", last_line);
        return std::nullopt;
    }

    const auto bm = str_utils::findElement<std::string>(str_utils::splitString(last_line, ' '), "bestmove");

    if (!bm.has_value()) {
        warn(bm.error(), last_line);
        return std::nullopt;
    }

    return *bm;
}
```

robertnurnberg:
Done. Changed the typo `Warning:` to `Warning;` in your snippet.

We will also have to take into account the slightly changed wording of the warning for the next fishtest worker update.

PR #1011, opened 2026-05-14
refactor: move mate score check into function

Disservin:


app/src/matchmaking/match/match.cpp
```diff
@@ -54,6 +54,47 @@ bool isFen(const std::string& line) { return line.find(';') == std::string::npos
     return {GameResultReason::NONE, GameResult::NONE};
 }
 
+// if both engines have made a move in this game, we perform a sanity check on their scores
+template <typename TScore>
+void checkMateScoreSignMismatch(const Player& us, const Player& them, const Board& board,
```
robertnurnberg, line 59: A tiny nit: the function compare the scores from them and us, with them having played the earlier move. That is why I had ordered everything in the order them, us. 

For example, as them played first, one would expect them the have a lower probability to have a mate score, and so it would make sense to check that one first. But also for the reader of the code it may be easier to leave everything in the order the moves were played in.

PR #1014, opened 2026-05-15
[fix] Do not compare scores from different games

robertnurnberg:
Re-organises the code a bit, including improving the comments.

~~One question: the line `if (data_.moves.size() > 1)` includes possible book moves having been "played"? If so, then we should find a better way to guard against comparing `usScore` from *this* game against `themScore` from the *previous* game. I.e. we would need access to a list of engine moves having been played this game.~~

So I checked the code and for pgn opening books master indeed does not catch this correctly. I pushed a fix that only counts the moves actually played by the engines.

### DIFF
--- a/app/src/matchmaking/match/match.cpp
+++ b/app/src/matchmaking/match/match.cpp
@@ -54,26 +54,24 @@ bool isFen(const std::string& line) { return line.find(';') == std::string::npos
     return {GameResultReason::NONE, GameResult::NONE};
 }
 
-// if both engines have made a move in this game, we perform a sanity check on their scores
-template <typename TScore>
+// emits a warning if both engines claim to have a proven win (same for loss)
 void checkMateScoreSignMismatch(const Player& them, const Player& us, const Board& board,
-                                const std::string& start_position, const MatchData& data, const TScore& usScore) {
+                                const std::string& start_position, const MatchData& data) {
     if (data.moves.size() <= 1) return;
 
-    const auto themScore = them.engine.lastScore();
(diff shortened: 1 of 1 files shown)


robertnurnberg:
I kept the bug fix as a separate commit. Let me know if you prefer a squashed PR.

Disservin:
don't think this a problem the book moves have no score so the code is just fine ?

robertnurnberg:
No, I fell foul of a similar situation in a the original pr. The function would recover the last score from them from the stdout. Which would be from the previous game.

robertnurnberg:
Unless we can recover the old them move from the move list, as you suggest....

robertnurnberg:
Let me try an alternative....

Disservin:
> No, I fell foul of a similar situation in a the original pr. The function would recover the last score from them from the stdout. Which would be from the previous game.

Well this is definitely more of a bug and that output should be cleared at the end of the game...

robertnurnberg:
So I pushed a more light-weight version now. This keeps all the logic regarding the moves list inside the function, like you had done in the refactor.

This could be merged as is to fix the bug. The fact that `lastScore()` may return the last score from the previous game, if the other engine has made a first move in the new game, can be interpreted as a bug, yes. But so far we do not rely on the engine output being cleared after a game is finished. If that is something easy to do, it should probably be done in a separate commit. (But would be above my pay grade.)

Disservin:
https://github.com/Disservin/fastchess/pull/1016

PR #1019, opened 2026-05-16
Use c'tor

robertnurnberg:


app/src/types/match_data.hpp
```diff
@@ -17,14 +17,14 @@ namespace fastchess {
...
-    MoveData(std::string _move, int64_t _elapsed_millis, int64_t _depth, int64_t _seldepth, uint64_t _nodes,
+    MoveData(std::string _move, Score _score, int64_t _elapsed_millis, int64_t _depth, int64_t _seldepth, uint64_t _nodes,
              bool _legal = true)
         : move(std::move(_move)),
           nodes(_nodes),
           elapsed_millis(_elapsed_millis),
           depth(_depth),
           seldepth(_seldepth),
-          legal(_legal) {}
+          legal(_legal) { score = _score; }
```
Disservin: ok but why did you add it like this and not like the other fields :D

robertnurnberg: Got some compilation error, I think connected to the variable being optional?

Disservin: idk works fine for me?

robertnurnberg: Oh! Maybe it was my wrong order then (I had tried to set score after move). Thanks!

PR #1023, opened 2026-05-29
logger-things

Disservin:


vondele:
btw, it might also be interesting to add to the losing on time message the amount the clock was overrun.

Disservin:
done in https://github.com/Disservin/fastchess/pull/1024

PR #1028, opened 2026-06-08
Exit nonzero on any warning using flag -strict

spinojara:
Solves #898.


app/src/core/logger/logger.hpp
```diff
@@ -131,6 +137,9 @@ class Logger {
 
         const std::lock_guard<std::mutex> lock(log_mutex_);
         std::visit([&](auto&& arg) { arg << fmt_message << std::flush; }, log_);
+
+        if (strict_ && level >= Level::WARN)
+            std::exit(1);
```
Disservin: i don't think this is correct since engines might not properly be killed.. better to set the global stop flags and interrupt flags add some code to main.cpp to return with exit code 1

Disservin: abnormal_termination, already does the correct thing in main.cpp

spinojara: You are totally right, forgot about possible running engines.

PR #1031, opened 2026-06-23
feat: better engine name defaults

gahtan-syarif:
Derive engine name from command when name not given, and add suffix to the engine name in cases where both engines have the same name.

solves https://github.com/Disservin/fastchess/issues/886

### DIFF
--- a/app/src/cli/sanitize.cpp
+++ b/app/src/cli/sanitize.cpp
@@ -5,7 +5,6 @@
 #include <thread>
 #include <utility>
 #include <vector>
-#include <filesystem>
 
 #include <core/filesystem/fd_limit.hpp>
 #include <core/filesystem/file_system.hpp>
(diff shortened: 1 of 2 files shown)


app/src/cli/sanitize.cpp
```diff
@@ -139,7 +135,17 @@ void validateEngine(EngineConfiguration& config) {
             throw fastchess_exception("Engine binary does not exist: " + enginePath.string());
         }
     }
+
+    if (config.name.empty()) {
+        // strip directory and extension
+        std::filesystem::path p(config.cmd);
```
Disservin, line 141: the other filesystem usages are guard with #ifndef NO_STD_FILESYSTEM

Disservin: nvm i just saw this is already in the same #ifdef

app/src/cli/sanitize.cpp
```diff
@@ -161,7 +167,7 @@ void sanitize(std::vector<EngineConfiguration>& configs) {
 
         for (std::size_t j = 0; j < i; j++) {
             if (configs[i].name == configs[j].name) {
-                throw fastchess_exception("Error: Engine with the same name are not allowed!: " + configs[i].name);
+                configs[i].name += "_" + std::to_string(i);
```
Disservin: this will always change the name, no ? if the user specified a name i wouldn't expect this to happen ?

gahtan-syarif: wouldnt it only change the name if user doesn't provide a name or gave the same name twice?

Disservin: sorry too early you are correct :D

Disservin:
can you adapt the test though?

gahtan-syarif:
> can you adapt the test though?

alright, will try 👍

gahtan-syarif:
added the test 👍

gahtan-syarif:
Hmm just noticed the code wouldnt really work with more than 2 engines, gotta rethink this

PR #1032, opened 2026-06-23
fix: default engine name when number of engines > 2

gahtan-syarif:
The original code is flawed when dealing with more than 2 engines that have the same name, this fixes it.

### DIFF
--- a/app/src/cli/sanitize.cpp
+++ b/app/src/cli/sanitize.cpp
@@ -166,16 +166,12 @@ void sanitize(std::vector<EngineConfiguration>& configs) {
         validateEngine(configs[i]);
     }
 
-    std::unordered_map<std::string, int> count;
-    for (auto& config : configs) {
-        count[config.name]++;
-    }
-
     std::unordered_map<std::string, int> seen;
     for (auto& config : configs) {
-        if (count[config.name] > 1) {
-            seen[config.name]++;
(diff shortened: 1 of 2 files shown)


gahtan-syarif:
modified the test to have number of engines > 2

app/src/cli/sanitize.cpp
```diff
@@ -164,12 +164,18 @@ void sanitize(std::vector<EngineConfiguration>& configs) {
...
-            if (configs[i].name == configs[j].name) {
-                configs[i].name += "_" + std::to_string(suffix++);
-            }
+    std::unordered_map<std::string, int> seen;
+    for (auto& config : configs) {
+        if (count[config.name] > 1) {
+            seen[config.name]++;
+            config.name += "_" + std::to_string(seen[config.name]);
         }
     }
```
Disservin:
```cpp
std::unordered_map<std::string, int> seen;

for (auto& config : configs) {
    int& n = seen[config.name];

    if (n++ > 0) {
        config.name += "_" + std::to_string(n);
    }
}
```

gahtan-syarif: ahh yeah thats way simpler

gahtan-syarif: fixed, thanks 👍

PR #1033, opened 2026-06-24
feat: default filename for pgnout and epdout

gahtan-syarif:
format looks like: `fastchess_20260624_080530.pgn/.epd`. i considered doing engine1_vs_engine2_timestamp but that doesnt really work out in a tournament of more than 2 engines and might exceed max character limits.


app/src/cli/cli.cpp
```diff
@@ -79,6 +79,16 @@ std::string concat(const std::vector<std::string>& params) {
     return str;
 }
 
+std::string getCurrentTimestampString() {
```
Disservin: just call the time::datetime function instead

gahtan-syarif: ahh ok, fixed 👍

PR #1044, opened 2026-07-20
feat: add suggestions for unrecognized command-line options

gahtan-syarif:
seems to work
image

### DIFF
--- a/app/src/cli/cli.hpp
+++ b/app/src/cli/cli.hpp
@@ -106,6 +106,51 @@ class OptionsParser {
                                   "\" with value \"" + std::string(value) + "\".");
     }
 
+    int levenshtein(std::string_view a, std::string_view b) {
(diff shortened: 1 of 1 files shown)


Disservin:
can this be improved by adding metaphone as well ?

gahtan-syarif:
> can this be improved by adding metaphone as well ?

probably, though i'd need to find the appropriate phonetic matching algo. though for this case, taking into account metaphones might not be as beneficial, since keyboard typos usually just adds or removes a letter (e.g. `rounds -> round` or `games -> gamees`) or switches letters (e.g. `rounds -> ruonds`) in which case using the levenshtein distance is enough

also git seems to just use levenshtein https://github.com/git/git/blob/master/levenshtein.c so its prob fine

Disservin:
LGTM

PR #1048, opened 2026-07-21
chore: update man

gahtan-syarif:
update man to include changes to default engine names as well as pgn and epd filenames


gahtan-syarif:
unsure why the snapshot tests failed

Disservin:
> unsure why the snapshot tests failed

well because the output obviously changed :D they need to be updated as well, document has different new line breaks so other parts also change

gahtan-syarif:
> > unsure why the snapshot tests failed
> 
> well because the output obviously changed :D they need to be updated as well, document has different new line breaks so other parts also change

i need to update this file? app/tests/snapshots/cli/help_intro.txt

Disservin:
i can also look at this later today, i anyway wanted to add more stuff regarding this

gahtan-syarif:
> i can also look at this later today, i anyway wanted to add more stuff regarding this

alright, couldnt figure out whats causing the diff on my end

PR #1061, opened 2026-08-03
feat: add -csvout to track llr/elo per interval

Carbecq:
## Summary

Adds a `-csvout` option that appends one row per interval to a CSV file, so a running SPRT can be plotted (llr vs games, elo vs games) e.g. in a spreadsheet.

- Follows the `-pgnout`/`-epdout` convention: `file` is optional (defaults to a timestamped name), `append`, plus `interval` (default: `-ratinginterval`) and `sep` for the column separator.
- Purely observational: it reads the same stats already computed for the console/interval output (`Stats`, `EloBase`, `SPRT::getLLR`) and does not affect the SPRT decision or the match scheduling.
- Columns: `iteration, games, pairs, llr, llr_lower, llr_upper, elo, elo_err, elo_lower, elo_upper, nelo, nelo_err, score, wins, losses, draws, LL, LD, WLDD, WD, WW`. The llr columns are left empty when `-sprt` is not used, and non-finite values (early nan/inf) are left blank rather than written as text.
- Not persisted in the JSON config, so it needs to be passed again when resuming with `-config` (same reasoning as other CLI-only knobs).

## Test plan

- [x] `make tests` — 165/165 test cases pass, including 3 new `-csvout` CLI parsing tests
- [x] Regenerated `man.hpp` (`make update-man`) and the `help.txt` CLI snapshot from the updated `man.md`
- [x] Manual run against a real SPRT match (self-play), confirmed the CSV rows match the console `printInterval` output exactly, including the row written when the SPRT bound is crossed


Carbecq:
Hi Disservin

  Just a gentle nudge on this one — no rush at all. The red X on this PR is a
  bit misleading: the CI never actually ran. The first run sat waiting for
  approval until it timed out, and the one from my latest push is waiting too.
  So there's no failing test hiding behind it.

  I've rebased it on current master and tidied up a couple of includes while I
  was there. Everything I could run locally is green.

Thanks for fastchess, by the way. I use it constantly.

PR #1064, opened 2026-08-15
fix: gauntlet Elo relative to base engine instead of aggregate score

### LINKED ISSUE #1007: Reporting Elo for Gauntlet Matches...
I've been using Fastchess to run some gauntlet matches. Here are some results from Fastchess where Juggernaut 1.55 is playing a range of opponents.

```
Rank Name                             Elo        +/-       nElo        +/-      Games      Score       Draw           Ptnml(0-2)
   1 Juggernaut 1.55               101.21      42.63     128.22      50.76        180      64.2%      40.0%   [5, 5, 36, 22, 22]
   2 CT800 v1.46                    53.88     134.87      57.12     133.55         26      57.7%      30.8%      [1, 3, 4, 1, 4]
   3 Fox Chess 1.0                 -81.65      75.76    -152.04     133.55         26      38.5%      69.2%      [2, 2, 9, 0, 0]
   4 Chal 1.4.0                    -95.90     115.14    -122.22     133.55         26      36.5%      46.2%      [3, 3, 6, 0, 1]
   5 Fatalii 0.10.0                -95.90     115.14    -122.22     133.55         26      36.5%      46.2%      [4, 1, 6, 2, 0]
   6 Juggernaut 1.40              -136.97     110.86    -199.37     139.00         24      31.2%      33.3%      [3, 4, 4, 1, 0]
   7 Reyna 1.0                    -190.85     103.99    -313.17     133.55         26      25.0%      30.8%      [4, 5, 4, 0, 0]
   8 Juggernaut 1.38              -190.85     131.22    -255.71     133.55         26      25.0%      23.1%      [5, 4, 3, 1, 0]
```
The Elo is being reported based on the Score %. This is quite misleading. For example, someone could naively think Juggernaut 1.55 is 291 Elo better than v1.38. Can I suggest that the Elo for a gauntlet match are always relative to the engine playing all games (in this case Juggernaut 1.38)?

— Steve

rehmansohail:
Fixes #1007

In gauntlet mode, per-engine Elo was calculated from aggregate score % 
against all opponents, which is misleading since non-seed engines never 
play each other. This made Elo differences between non-seed engines hard 
to interpret (e.g. it looked like one was ~290 Elo better than another 
purely based on overall score).

@Disservin


Disservin:
what's cutechess doing?

rehmansohail:
> what's cutechess doing?

Cutechess is considering all the matches played (even the ones which non-base engines play with other non-base engines). For a metric which measures the benchmark against a base engine, these additional matches (amongst non-base engines) may lead to incorrect Elo values

